PR #7, opened 2020-06-12
Upgrade JavaScript dependencies

darabos:
After pushing to the new repo I immediately got a couple Dependabot alerts for outdated dependencies. I've used `yarn upgrade --latest` to bump everything to a more recent version. After fixing a few new ESLint warnings everything seems fine!


web/app/scripts/workspace/workspace-wrapper.js
```diff
@@ -116,7 +116,7 @@ angular.module('biggraph')
           const dst = this.boxes[i];
           const inputs = dst.instance.inputs;
           for (let inputName in inputs) {
-            if (inputs.hasOwnProperty(inputName)) {
+            if (Object.prototype.hasOwnProperty.call(inputs, inputName)) {
```
xandrew-lynx, line 119: Why is this necessary? The old version certainly looked better... :)

web/app/scripts/workspace/workspace-wrapper.js
```diff
@@ -386,11 +386,11 @@ angular.module('biggraph')
           const oldBox = boxes[i];
           const newBox = mapping[oldBox.id];
           for (let key in oldBox.inputs) {
-            if (!oldBox.inputs.hasOwnProperty(key)) {
-              break;
+            if (!Object.prototype.hasOwnProperty.call(oldBox.inputs, key)) {
+              continue;
```
xandrew-lynx, line 390: Was this a bug? I also don't understand what this if is doing... Isn't the above for only iterates over "own" properties? And if not, how come this never caused any issue?

web/app/scripts/workspace/workspace-wrapper.js
darabos: > Why is this necessary? The old version certainly looked better... :)

Yeah, this looks so bad. Read more on the [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins) ESLint rule and the [Iterate through object properties](https://stackoverflow.com/questions/8312459/iterate-through-object-properties) Stack Overflow question.

I think the short story is that the object may have stuff we want to ignore from the Object prototype, so we need the check. But it also may _not_ have stuff from that prototype, in which case `inputs.hasOwnProperty` would be undefined.

Now that I look more closely, it looks like iterating over [`Object.entries(inputs)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) would look much nicer. I'll switch to that.

web/app/scripts/workspace/workspace-wrapper.js
darabos: > Was this a bug? I also don't understand what this if is doing... Isn't the above for only iterates over "own" properties?

I think it was a bug to `break` on the first property from the prototype. Depending on the ordering we could have ignored real entries because they came after a prototype property.

> And if not, how come this never caused any issue?

The whole thing is a theater! These are plain JavaScript objects that don't have any prototype properties. `x = {a: 1, b: 2}; for (let k in x) { console.log(k); }` just prints `a` and `b`.

darabos:
Sorry, I'll delay this until after the release. Turns out when I quickly tested this I tested in the wrong directory. 😓  The upgrades did break a few things. I've fixed them mostly now (will push soon), but I don't see a reason to make the release even more risky. We can upgrade these in the second release.

xandrew-lynx:
Sounds good to me!

PR #10, opened 2020-06-15
Fixes from testing wizards

darabos:
I had somehow missed these earlier. But I think it's working fairly well now.


web/app/styles/graph-view.css
```diff
@@ -2,18 +2,17 @@ svg.graph-view {
...
   --c: #39bcf340;
-  /* Because of the !important background above, these need to be !important too. */
   background-image: linear-gradient(
     45deg, transparent 25%, var(--c) 25%, var(--c) 26%,
     transparent 26%, transparent 27%, var(--c) 27%, var(--c) 50%,
     transparent 50%, transparent 75%, var(--c) 75%, var(--c) 76%,
-    transparent 76%, transparent 77%, var(--c) 77%, var(--c)) !important;
-  background-size: 200px 200px !important;
+    transparent 76%, transparent 77%, var(--c) 77%, var(--c));
+  background-size: 200px 200px;
```
olahg, line 15: Okay, I guess this here: https://tosbourn.com/firefox-honours-important-in-css-animations-no-one-else-seems-to/ has to do with it. But then it will not show in print, right?

darabos: > Okay, I guess this here: https://tosbourn.com/firefox-honours-important-in-css-animations-no-one-else-seems-to/ has to do with it.

I think it's just a bug. https://codepen.io/P1N2O/pen/pyBNzX works for me. But if I add `!important` to every rule, it doesn't work. Like the article says, I would expect that adding `!important` to the keyframes would make the animation just as important as the base rule.

> But then it will not show in print, right?

Fewer users access LynxKite through a printer than through a screen, so I thought it's a good trade. 😅   But I've checked it now, and the background was already missing from the print version before this change. So... not sure whether to be happy or sad about that. It doesn't look too bad though:

image

PR #12, opened 2020-06-16
Move dependency licenses into subfolder to avoid confusing Licensee

darabos:
GitHub uses https://github.com/licensee/licensee to show the license of the repo. Our dependency license files confuse it. Local testing suggests that moving them into a directory fixes this. The repo looks a bit tidier too!


darabos:
It worked!

image

PR #24, opened 2020-06-22
Update license in Python package

darabos:
I've pushed this to PyPI: https://pypi.org/project/lynxkite-python-api/

### DIFF
--- a/python/documentation/Introduction.ipynb
+++ b/python/documentation/Introduction.ipynb
@@ -68,7 +68,7 @@
     "If you are building your own LynxKite setup, you will have to install the Python package:\n",
     "\n",
     "```\n",
-    "pip install lynxkite-python-api\n",
+    "pip install lynxkite-client\n",
     "```\n",
     "\n",
     "(Requires Python 3.6 or newer.)\n",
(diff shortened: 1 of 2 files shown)


darabos:
> Maybe it's too late, but let's consider. I think the name of this package is suboptimal. It's long and the word "python" is redundant when we are naming a python package. "lynxkite" or "lynxkite-api" would be better IMHO. A rename would also create a clear cut between the "old" proprietary stuff and the new shiny OS version. So what do you think about a rename? If it's too much trouble now, then it's fine of course.

I was pondering that too. I've already pushed this to `lynxkite-python-api`, but I think it's fine to just delete that and create the new package. Maybe `lynxkite-client` would be the most descriptive name?

xandrew-lynx:
Yeah, I like lynxkite-client!

darabos:
All set:
- https://pypi.org/project/lynxkite-client/
- https://pypi.org/project/lynxkite-python-api/ (Doesn't contain any files, but depends on `lynxkite-client`, so `pip install lynxkite-python-api` still works.)

PR #28, opened 2020-06-25
Remove ng-annotate and ng-filesort

darabos:
`ng-annotate` is no longer maintained. Its replacement is based on Babel which we don't use. It brings in a lot of obsolete dependencies, and it forces us to avoid using some new JavaScript syntax.

And, apparently, it can be replaced by a regular expression.

The processed files are identical before and after the second commit.


xandrew:
Sorry, it's very hard to see what really is going on here with all the formatting changes. Is it possible to split those out into a separate commit, or alternatively, give some "reviewers' guide" to this PR?

darabos:
> Sorry, it's very hard to see what really is going on here with all the formatting changes. Is it possible to split those out into a separate commit, or alternatively, give some "reviewers' guide" to this PR?

If you look at the first commit, it's just adding `angular.module('biggraph')` everywhere instead of chaining. This results in some whitespace changes too, enforced by ESLint. You're right that it would have been better to have these in separate commits, sorry about that. But if you only look at this commit, maybe it's tolerable? GitHub is pretty helpful with the highlighting.

xandrew:
Just checking, what's up with this PR? Merge after some conflict resolution?

darabos:
> Just checking, what's up with this PR? Merge after some conflict resolution?

The tests are green. It can only mean MERGE!!!

PR #36, opened 2020-07-01
Use Go modules in Sphynx

darabos:
This is originally to fix https://github.com/xitongsys/parquet-go/issues/292. But it's just much nicer overall. It makes the build reproducible. It speeds up the build too, because we don't have to upgrade packages all the time.


sphynx/go.sum
```diff
@@ -0,0 +1,351 @@
...
+gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
+honnef.co/go/tools v0.0.0-20190102054323-c2f93a96b099/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
+honnef.co/go/tools v0.0.0-20190106161140-3f1c8253044a/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
+honnef.co/go/tools v0.0.0-20190418001031-e561f6794a2a/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
+honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
+honnef.co/go/tools v0.0.1-2019.2.3/go.mod h1:a3bituU0lyd329TUQxRnasdCoJDkEUEAqEt0JzvZhAg=
+honnef.co/go/tools v0.0.1-2020.1.3/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9vFzvIQ3k=
+rsc.io/binaryregexp v0.2.0/go.mod h1:qTv7/COck+e2FymRvadv62gMdZztPaShugOCi3I+8D8=
+rsc.io/quote/v3 v3.1.0/go.mod h1:yEA65RcK8LyAZtP9Kv3t0HmxON59tX3rD+tICJqUlj0=
+rsc.io/sampler v1.3.0/go.mod h1:T1hPZKmBbMNahiBKFy5HrXp6adAjACjK9JXDnKaTXpA=
```
xandrew, line 351: What is this file meant to do? Why a lot of different version from some packages? Is this meant to be under git?

xandrew:
Highly synced team. Was just looking at this. :) But yeah, it looks good, just some novice question.

sphynx/go.sum
darabos: > What is this file meant to do? Why a lot of different version from some packages? Is this meant to be under git?

Yes. This is like `yarn.lock`. It's automatically managed to track the exact versions that get pull in as dependencies of dependencies defined in `go.mod`.

PR #40, opened 2020-07-01
Remove some bits about user authentication from the admin manual

darabos:
This whole admin manual belongs to a museum I think. What if we just deleted the whole thing?


darabos:
Thanks for the quick reviews! I'll roll a release today.

> I don't know, I'm not necessarily against deleting this. But you read it through now: any good information not yet elswhere? Maybe those should be moved to a new location then. Where? Normal manual? Website? Or not useful information left here?

The `kiterc` section is necessary and fairly clear. But some parts are somewhat confusing (_"As a rule of thumb it is recommended to plan at least with 10 times of the storage capacity of the original data. So if you have 100 GB data to analyse in LynxKite have at least 500 GB set aside for permanent storage."_ [[1]](https://lynxkite.com/docs/latest/admin-manual/#_sizing_the_cluster)) or clearly outdated (_"LynxKite runs on any modern Linux distributions. We recommend using Ubuntu Server 14.04 LTS or CentOS/Redhat 6.7, because LynxKite has been tested on those systems."_ [[2]](https://lynxkite.com/docs/latest/admin-manual/#_recommendations)).

On one hand it's nice that we have a section on cluster sizing that we can point people to if they need it. On the other hand I would really hope nobody reads these sections. 😸  I think these were mostly written with an internal audience in mind. For the open-source version we could just delete them. If someone needs to size a cluster, it depends on so many factors anyway. It's best if they just ask us.

I'll make another PR with a lot of deletions and we can discuss what to keep there.

PR #45, opened 2020-07-03
Use protoc-gen-go-grpc

darabos:
The same thing could actually also work with the old `protoc-gen-go`. But it looks like they just happen to be changing the whole thing, so this is also an upgrade.

[`github.com/golang/protobuf`](https://github.com/golang/protobuf) is the old one. [`google.golang.org/protobuf`](https://pkg.go.dev/mod/google.golang.org/protobuf) is the new one.

In the old thing the GRPC generator was a plugin of the Go `protoc` plugin. In the new thing it is a separate `protoc` plugin. (See https://github.com/grpc/grpc-go/pull/3453.)

The API is largely unchanged, but now requires this UnimplementedServer thing. (See https://github.com/grpc/grpc-go/pull/3657.)

We still depend on the old library in `go.mod`. This is because the generated code still imports the old library. Why does the new generator generate code that imports the old library?! I think either I've messed it up somehow, or they haven't updated the generated imports yet. Either way, everything works. We can figure it out later.

### DIFF
--- a/sphynx/.gitignore
+++ b/sphynx/.gitignore
@@ -1,6 +1,4 @@
 proto/*.pb.go
-protoc/*
 *.zip
-*.exe
 *.pem
 /.build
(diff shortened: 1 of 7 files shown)


darabos:
> Thanks! It does work for me as well!

Unexpected but welcome news. 😃

PR #47, opened 2020-07-07
Add some useful GPL dependencies

### LINKED ISSUE #46: Include GPL JDBC drivers
We used to not include GPL JDBC drivers (notably MySQL's) in LynxKite because it wasn't itself a GPL licensed software. But now it is! Let's add MySQL and whatever else!

darabos:
Resolves #46.


darabos:
> I guess we need a corresponding PR in lynxkite-enterprise removing these?

I think so.

PR #51, opened 2020-07-11
Filter with SQL box

darabos:
Why am I not at the beach? 🏖️ 

image

TODO:
- Add a test.
- Add Sphynx implementations of these operations.


app/com/lynxanalytics/biggraph/graph_operations/SQLFiltering.scala
```diff
@@ -0,0 +1,58 @@
...
+}
+
+object EmbeddingByID extends OpFromJson {
+  class Input extends MagicInputSignature {
+    val original = vertexSet
+    val filtered = vertexSet
+  }
+  class Output(implicit instance: MetaGraphOperationInstance, inputs: Input) extends MagicOutput(instance) {
+    val identity = edgeBundle(inputs.filtered.entity, inputs.original.entity, EdgeBundleProperties.embedding)
+  }
```
darabos, line 41: Maybe we have this somewhere already? Seems like such an obvious operation.

xandrew: It's basically VertexSetIntersection. Two differences: that one does some work to really just keep the common ids and the result is partitioned as the original in that case. I'm not hundred percent sure whether these are good or bad. :) But we use that for attribute filters, so maybe safest to use it here as well?

xandrew: An alternative would be if a single operation would do the whole job. Input: a VertexSet and an id table. Output: the new vertex set and its embedding. I don't see how creating a VertexSet out of nothing from ids will be useful on its own. And then it's nicer not to create the embedding after the fact. WDYT?

xandrew:
Cool stuff! By large, looks great, added some technical comments!

app/com/lynxanalytics/biggraph/graph_operations/SQLFiltering.scala
xandrew: During breakfast I thought of a better argument for using a single operation than "It feels nicer to me". In the current two operations setup, a new partitioner is pulled out of thin air, it will be always different than the original, so everything will have to be repartitioned. If this new partitioner is very different, then it's fine. But if it's just slightly different, e.g. we went from 1M to 900k vertices, then it would be better to keep the original partitioner. If on the other hand we use the partitioner of the original VS, then exactly the right thing will happen: if the difference is small, we keep it. If the difference is big then the data manager will realize and repartition.

darabos:
> * Add a test.

Done.

> * Add Sphynx implementations of these operations.

We don't have support for Tables in Sphynx. It would probably not be hard to add them, but anyway the SQL runs on Spark, so there's not to be gained here.

app/com/lynxanalytics/biggraph/frontend_operations/WorkflowOperations.scala
```diff
@@ -501,6 +501,83 @@ class WorkflowOperations(env: SparkFreeEnvironment) extends ProjectOperations(en
...
+    def enabled = FEStatus.enabled
+    override def getOutputs() = {
+      params.validate()
+      if (input.isProject) {
+        val project = input.project
+        def tables = project.viewer.getProtoTables.toMap
+        val before = project.rootEditor.viewer
+        if (params("vertex_filter").trim.nonEmpty) {
+          project.vertexAttributes.updateEntityMap(
+            project.vertexAttributes.iterator.toMap + ("!id" -> project.vertexSet.idAttribute))
```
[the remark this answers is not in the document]
darabos: > Why can't we just use newVertexAttribute for this?

Because it would convert the IDs to Double. 😓

app/com/lynxanalytics/biggraph/frontend_operations/WorkflowOperations.scala
```diff
@@ -501,6 +501,83 @@ class WorkflowOperations(env: SparkFreeEnvironment) extends ProjectOperations(en
...
+          val vf = graph_operations.ExecuteSQL.run(
+            "select `!id` from vertices where " + params("vertex_filter"), tables)
+          project.vertexAttributes.updateEntityMap(
```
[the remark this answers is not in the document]
darabos: > This back and forth editing of the project is a bit hacky. Why not create a new copy of the project for the generation of the table instead of editing "the" project?

Done.

app/com/lynxanalytics/biggraph/frontend_operations/WorkflowOperations.scala
```diff
@@ -501,6 +501,83 @@ class WorkflowOperations(env: SparkFreeEnvironment) extends ProjectOperations(en
     registerSQLOp(s"SQL$inputs", numbers.take(inputs))
   }
 
+  registerOp("Filter with SQL", "filter", category, List("input"), List("output"), new SmartOperation(_) {
+    private val input = context.inputs("input")
+    if (input.isProject) {
+      params ++= List(
+        Code("vertex_filter", "Vertex filter", language = "sql"),
+        Code("edge_filter", "Edge filter", language = "sql"))
+    } else {
```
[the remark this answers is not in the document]
darabos: > Let's explicitly check for table kind here and return an error otherwise (say, for visualization). (But it's really cool that this works for tables as well! :) )

Done.

app/com/lynxanalytics/biggraph/graph_operations/SQLFiltering.scala
darabos: > During breakfast I thought of a better argument for using a single operation than "It feels nicer to me".

Good points, thanks. Done.

darabos:
Wow, I sent this a month ago? Felt like just a few days. 👀

app/com/lynxanalytics/biggraph/graph_operations/FilterByTable.scala
```diff
@@ -0,0 +1,36 @@
...
+  override def toJson = Json.obj("column" -> column)
+  def execute(
+    inputDatas: DataSet,
+    o: Output,
+    output: OutputBuilder,
+    rc: RuntimeContext): Unit = {
+    implicit val id = inputDatas
+    val partitioner = inputs.vs.rdd.partitioner.get
+    val ids = inputs.t.data.df.select(column).rdd.map(_.getLong(0))
+    val vs = ids.map((_, ())).sortUnique(partitioner)
```
xandrew, line 32: I think we should do a sorted join with the original vertices here. Otherwise bad things happen if the table has ids that are not originally part of the vertex set.

darabos: > I think we should do a sorted join with the original vertices here. Otherwise bad things happen if the table has ids that are not originally part of the vertex set.

Oh, good point. Done.

darabos:
Thanks for the quick reviews! I will make sure the tests pass on stuff and then merge them while fixing the CHANGELOG conflicts. 😅

PR #59, opened 2020-07-27
Fix `proto_compile.sh` on Mac

bramrodenburg:
**Description**
Building LynxKite with `make` doesn't work on a Mac, because the Linux version version of `protobuf` and `protoc-gen-grpc-java` are downloaded by default in `sphynx/proto_compile.sh`. In this pull request I'm addressing this by checking the operating system and downloading the correct version (either `osx` or `linux`).

### DIFF
--- a/sphynx/proto_compile.sh
+++ b/sphynx/proto_compile.sh
@@ -5,10 +5,17 @@
 cd $(dirname $0)
 . sphynx_common.sh
 
+# Check operating system for protobuc and grpc.
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  OS_NAME="osx"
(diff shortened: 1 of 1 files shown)


darabos:
The frontend test failure looks like a new kind of flakiness:
```
    Failed: unknown error: unhandled inspector error: {"code":-32000,"message":"Inspected target navigated or closed"}

```
I'll look closer if it comes back often.

bramrodenburg:
> Cool thanks! This is the only change needed for Mac development? 👀

Well after that change I could at least build everything with `make`. I'm still running into some errors when trying to run LynxKite on my Mac, but still need to look into that a bit more.

PR #60, opened 2020-07-28
Documentation for LynxKite Operations in Python

bramrodenburg:
**Description**
In this pull request I'm trying to finish what I started "quite" some time ago. That is, I'm trying to improve the LynxKite documentation for the Python API.

On a high level, I have implemented a script that can convert the `.asciidoc` operations documentation to Sphinx compatible Python descriptions for the LynxKite operations that a user can call on `LynxKite`. These descriptions are stored in the file called `operations.py`. As a result, a user can now call `help(...)` on a given operation and see some meaningful documentation. 

E.g.:
```python
import lynx.kite
lk = lynx.kite.LynxKite(address='http://localhost:2200/')
help(lk.createExampleGraph)
```
now shows:
```
Help on function createExampleGraph in module lynx.kite:

createExampleGraph(*args, **kwargs)
    Creates small test graph with 4 people and 4 edges between them.
```

I also added a `build.sh` script for generating the `operations.py` and Sphinx documentation in one go. Lastly, I fixed a parameter descriptions in a few `.asciidoc` files.

**Screenshots of how the Sphinx documentation looks like**
image


python/documentation/create_operations_doc.py
```diff
@@ -0,0 +1,71 @@
+''' Script for generating Python documentation from the ASCIIDOC documentation.
+
+
+
```
darabos, line 4: Maybe one or two empty lines would be enough. 😄

python/documentation/create_operations_doc.py
```diff
@@ -0,0 +1,71 @@
...
+  body = textwrap.indent(re.search(BODY_REGEX, content).groups()[0].strip(), '  ')
+  params = re.findall(PARAMS_REGEX, content)
+  params = [] if not params else params
+
+  params_text = ''
+  for name, desc in params:
+    params_text += textwrap.indent(f':param {name}: {textwrap.indent(desc.strip(), "  ")}\n', '  ')
+  params_text = params_text.rstrip()
+  return f'''
+def {operation_name}(*args, **kwargs):
```
darabos, line 45: Since you already know the parameter names, maybe it would not be much more trouble to put them in the argument list. If it would allow autocompletion for the parameter names when used in a notebook that would be super convenient.

python/documentation/create_operations_doc.py
```diff
@@ -0,0 +1,71 @@
...
+def get_python_name(file_name):
+  '''Converts the file name into the corresponding function name.'''
+  return _python_name(file_name[:-len('.asciidoc')].replace('-', ' '))
+
+
+def generate_documentation(inpath, outpath):
+  with open(outpath, 'w') as fh:
+    fh.write(HEADER)
+
+    for path in glob.glob(os.path.join(ASCIIDOC_OPERATIONS_PATH, '*.asciidoc')):
```
darabos, line 62: Is the glob order predictable? Maybe better to call `sorted()` on it so we won't have spurious changes in `operations.py`.

python/documentation/create_operations_doc.py
bramrodenburg: I was considering this myself as well. However, the problem that I see is that some parameters are `optional` (with a default value) whereas others are `required`. I'm not always able to infer this from the asciidocs. Unless I'm wrong here, do you think it's worth making the tradeoff of adding the parameters, even thought it comes at the cost of correctness?

darabos: I don't think you can figure out the default values, so it would always appear as required, right? But what exactly would be the effect?
- If the documentation says `pageRank(weight, iterations)`, that's clearly better than if it says `pageRank(*args, **kwargs)`. The text for the parameter may clarify whether there is a default.
- If I type `pageRank(wei<TAB>` in a notebook and it autocompletes that's fantastic.

Would there be any downside, do you think?

bramrodenburg: I'm convinced. I'll implement it.

python/documentation/create_operations_doc.py
bramrodenburg: Done.

python/documentation/create_operations_doc.py
bramrodenburg: Done.

python/documentation/create_operations_doc.py
bramrodenburg: Fixed.

bramrodenburg:
@darabos I think it's good enough to merge right now. I changed the formatting in a few `.asciidoc` files, because that made the parsing of the files much easier for me. I hope this is ok. 

If you want, you can checkout the generated Sphinx documentation yourself to see whether it's satisfactory enough.

darabos:
There may be a special place in the afterlife for people who convert AsciiDoc to RST with regular expressions. 😄  Hopefully it's a good place. Thanks!

PR #70, opened 2020-08-09
Added state.vertices, state.edges, etc.

bramrodenburg:
**Description**
Added state properties for easily accessing `vertices`, `edges`, `graph_attributes`,  `edge_attributes` and `segmentations` in the Python Remote API. 

You can now do things such as:

```python
g = lk.createExampleGraph()
g.vertices
g.edges
g.edge_attributes
g.graph_attributes

g2 = g.findConnectedComponents(name="example")
g.segmentations["example"]
```

### DIFF
--- a/python/remote_api/src/lynx/kite.py
+++ b/python/remote_api/src/lynx/kite.py
@@ -904,16 +904,15 @@ class State:
     '''Returns the edge attributes as a table.'''
     return self._select_all('edge_attributes')
 
-  @property
-  def segmentations(self):
-    '''Access a given segmentation.
+  def segmentations(self, name: str) -> 'SingleOutputAtomicBox':
+    '''Access a given segmentation by name.
 
     Example usage:
     ```
     graph = lk.createExampleGraph().findConnectedComponents(name='seg1')
-    segmentation = graph.segmentations['seg1']
+    segmentation = graph.segmentations('seg1')
     '''
-    return SegmentationPlaceholder(self)
+    return self.takeSegmentationAsBaseGraph(apply_to_graph='.' + name)
 
   def _select_all(self, table):
     return self.sql(f'select * from `{table}`')
@@ -1012,16 +1011,6 @@ class Placeholder:
 
   def __init__(self, value=None) -> None:
     self.value = value
-
-
-class SegmentationPlaceholder:
-  '''Placeholder for a segmentation.'''
-
-  def __init__(self, state: 'State'):
-    self.state = state
-
-  def __getitem__(self, name):
-    return self.state.takeSegmentationAsBaseGraph(apply_to_graph='.' + name)
 
 
 def _fn_id(fn: Callable):
--- a/python/remote_api/src/lynx/kite.py
+++ b/python/remote_api/src/lynx/kite.py
@@ -904,8 +904,8 @@ class State:
     '''Returns the edge attributes as a table.'''
     return self._select_all('edge_attributes')
 
-  def segmentations(self, name: str) -> 'SingleOutputAtomicBox':
-    '''Access a given segmentation by name.
+  def segmentation(self, name: str) -> 'SingleOutputAtomicBox':
+    '''Returns the named segmentation as a base project.
 
     Example usage:
     ```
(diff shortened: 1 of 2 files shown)


python/remote_api/src/lynx/kite.py
```diff
@@ -884,6 +884,40 @@ def f(**kwargs):
...
+    return self._select_all('edge_attributes')
+
+  @property
+  def segmentations(self):
+    '''Access a given segmentation.
+
+    Example usage:
+    ```
+    graph = lk.createExampleGraph().findConnectedComponents(name='seg1')
+    segmentation = graph.segmentations['seg1']
```
darabos, line 914: I think this syntax makes it look like `graph.segmentations` is a dict. So a user may try `for s in graph.segmentations` or stuff like that which won't work. For this reason I think a simple function call is more honest and not any more verbose.

```suggestion
    segmentation = graph.segmentation('seg1')
```

python/remote_api/src/lynx/kite.py
```diff
@@ -884,6 +884,40 @@ def f(**kwargs):
...
+    return self._select_all('graph_attributes')
+
+  @property
+  def edge_attributes(self) -> 'SingleOutputAtomicBox':
+    '''Returns the edge attributes as a table.'''
+    return self._select_all('edge_attributes')
+
+  @property
+  def segmentations(self):
+    '''Access a given segmentation.
```
darabos, line 909: Maybe mention that it uses a _Use segmentation as base graph_ box. That would clarify what kind of access it gives. (E.g. no belongs-to edges.)

python/remote_api/src/lynx/kite.py
bramrodenburg: Done. Is this approximately what you had in mind?

python/remote_api/src/lynx/kite.py
bramrodenburg: Given that I changed the `graph.segmentations` to a function call that now contains a one-liner that shows the `takeSegmentationAsBaseGraph` method, I think this is not really required anymore?

python/remote_api/src/lynx/kite.py
```diff
@@ -884,6 +884,39 @@ def f(**kwargs):
...
+  def graph_attributes(self) -> 'SingleOutputAtomicBox':
+    '''Returns the graph attributes as a table.'''
+    return self._select_all('graph_attributes')
+
+  @property
+  def edge_attributes(self) -> 'SingleOutputAtomicBox':
+    '''Returns the edge attributes as a table.'''
+    return self._select_all('edge_attributes')
+
+  def segmentations(self, name: str) -> 'SingleOutputAtomicBox':
```
darabos, line 907: Sorry, just minor tweaks. I'd do them but I'm not sure if I can commit to your fork.
```suggestion
  def segmentation(self, name: str) -> 'SingleOutputAtomicBox':
```

It returns a single segmentation so let's call it `segmentation()`.

python/remote_api/src/lynx/kite.py
```diff
@@ -884,6 +884,39 @@ def f(**kwargs):
...
+    '''Access a given segmentation by name.
```
darabos, line 908:
```suggestion
    '''Returns the named segmentation as a base project.
```

It's concise but specific enough to hint at the box. What do you think?

python/remote_api/src/lynx/kite.py
darabos: > Given that I changed the `graph.segmentations` to a function call that now contains a one-liner that shows the `takeSegmentationAsBaseGraph` method, I think this is not really required anymore?

The documentation shouldn't assume that the user has read the source code. 😄

python/remote_api/src/lynx/kite.py
bramrodenburg: I agree! Changed it.

python/remote_api/src/lynx/kite.py
bramrodenburg: Done.

darabos:
Not sure what's wrong with the frontend test but I'll look at it later. Thanks!

PR #78, opened 2020-08-12
Allow disabling suffixes in aggregations

### LINKED ISSUE #67: Make aggregation suffixes optional in merges
I almost always find it annoying that a _sum suffix is added. Of course, it's necessary if the user selects multiple aggregation for the same attribute, but otherwise the semantics is more often then not that the user tells the system the "natural" way to aggregate an attribute.

Maybe a checkbox or something? If the user opts for no suffix then it's an error to choose more than one aggregation per field.

darabos:
Resolves #67.

### DIFF
--- a/app/com/lynxanalytics/biggraph/frontend_operations/AttributePropagationOperations.scala
+++ b/app/com/lynxanalytics/biggraph/frontend_operations/AttributePropagationOperations.scala
@@ -17,11 +17,11 @@ class AttributePropagationOperations(env: SparkFreeEnvironment) extends ProjectO
     def addSegmentationParameters = params ++= aggregateParams(parent.vertexAttributes)
     def enabled = project.assertSegmentation
     def apply() = {
-      for ((attr, choice) <- parseAggregateParams(params)) {
+      for ((attr, choice, name) <- parseAggregateParams(params)) {
         val result = aggregateViaConnection(
           seg.belongsTo,
           AttributeWithLocalAggregator(parent.vertexAttributes(attr), choice))
-        project.newVertexAttribute(s"${attr}_${choice}", result)
+        project.newVertexAttribute(name, result)
       }
     }
     override def cleanParametersImpl(params: Map[String, String]) = cleanAggregateParams(params)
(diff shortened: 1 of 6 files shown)


app/com/lynxanalytics/biggraph/frontend_operations/Operations.scala
```diff
@@ -178,55 +178,63 @@ abstract class ProjectOperations(env: SparkFreeEnvironment) extends OperationReg
...
-        val options = if (attr.is[Double]) {
-          if (weighted) { // At the moment all weighted aggregators are global.
-            FEOption.list("weighted_average", "by_max_weight", "by_min_weight", "weighted_sum")
-          } else if (needsGlobal) {
-            FEOption.list(
-              "average", "count", "count_distinct", "count_most_common", "first", "max", "min", "most_common",
-              "std_deviation", "sum")
-
+    Choice("add_suffix", "Add suffixes to attribute names", options = FEOption.yesno) +:
+      sortedAttrs.toList.map {
```
darabos, line 202: I didn't change this, it just gets indented by Scalariform. I've only added the line above.

app/com/lynxanalytics/biggraph/frontend_operations/AttributePropagationOperations.scala
```diff
@@ -41,11 +41,11 @@ class AttributePropagationOperations(env: SparkFreeEnvironment) extends ProjectO
       def apply() = {
         val weightName = params("weight")
         val weight = parent.vertexAttributes(weightName).runtimeSafeCast[Double]
-        for ((attr, choice) <- parseAggregateParams(params)) {
+        for ((attr, choice, name) <- parseAggregateParams(params, weight = weightName)) {
```
xandrew, line 44: Looking at this code it occurred to me that it seems inconsistent that we don't allow specifying a prefix in case of the aggregations towards segmentations. Not that I ever really missed it in practice, but maybe we should just add a prefix parameter in these two operations as well, defaulting to empty string?

xandrew: If you do this, you could avoid the prefix parameter for `parseAggregateParams`. Instead, it can take it from params. And the parameter could be added by `aggregateParams`. It would take a new argument with the default value for prefix, defaulting to `""`.

But all this is not too important, if there is some complication I don't see then it's fine to ignore this.

darabos: > If you do this, you could avoid the prefix parameter for `parseAggregateParams`. Instead, it can take it from params. And the parameter could be added by `aggregateParams`. It would take a new argument with the default value for prefix, defaulting to `""`.

Very nice, thanks! Done.

PR #80, opened 2020-08-13
Aggregation for vectors

### LINKED ISSUE #56: Add aggregations for vectors
For lat/long vectors it totally makes sense to take the average. Some other aggregations may also be useful on some datasets. (E.g. element-wise min/max.)

darabos:
Resolves #56. We didn't have very thorough testing for this so I tried to improve that a bit.

### DIFF
--- a/app/com/lynxanalytics/biggraph/controllers/FEAggregations.scala
+++ b/app/com/lynxanalytics/biggraph/controllers/FEAggregations.scala
@@ -46,26 +46,39 @@ object AttributeWithAggregator {
     attr: Attribute[T], choice: String)(
     implicit
     manager: MetaGraphManager): AttributeWithAggregator[_, _, _] = {
-
+    def d = attr.runtimeSafeCast[Double]
+    def v = attr.runtimeSafeCast[Vector[Double]]
     choice match {
       case "sum" =>
-        AttributeWithAggregator(attr.runtimeSafeCast[Double], graph_operations.Aggregator.Sum())
+        AttributeWithAggregator(d, graph_operations.Aggregator.Sum())
       case "count" => AttributeWithAggregator(attr, graph_operations.Aggregator.Count[T]())
       case "min" =>
-        AttributeWithAggregator(attr.runtimeSafeCast[Double], graph_operations.Aggregator.Min())
+        AttributeWithAggregator(d, graph_operations.Aggregator.Min())
       case "max" =>
-        AttributeWithAggregator(attr.runtimeSafeCast[Double], graph_operations.Aggregator.Max())
+        AttributeWithAggregator(d, graph_operations.Aggregator.Max())
       case "average" => AttributeWithAggregator(
-        attr.runtimeSafeCast[Double], graph_operations.Aggregator.Average())
+        d, graph_operations.Aggregator.Average())
       case "first" => AttributeWithAggregator(attr, graph_operations.Aggregator.First[T]())
       case "std_deviation" => AttributeWithAggregator(
-        attr.runtimeSafeCast[Double], graph_operations.Aggregator.StdDev())
+        d, graph_operations.Aggregator.StdDev())
       case "most_common" =>
         AttributeWithAggregator(attr, graph_operations.Aggregator.MostCommon[T]())
       case "count_most_common" =>
         AttributeWithAggregator(attr, graph_operations.Aggregator.CountMostCommon[T]())
       case "count_distinct" =>
         AttributeWithAggregator(attr, graph_operations.Aggregator.CountDistinct[T]())
+      case "elementwise_sum" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseSum())
+      case "elementwise_min" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseMin())
+      case "elementwise_max" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseMax())
+      case "elementwise_average" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseAverage())
+      case "elementwise_std_deviation" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseStdDev())
+      case "concatenate" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.Concatenate[Double]())
     }
   }
 }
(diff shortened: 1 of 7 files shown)


app/com/lynxanalytics/biggraph/controllers/FEAggregations.scala
```diff
@@ -46,26 +46,39 @@ object AttributeWithAggregator {
...
+      case "elementwise_min" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseMin())
+      case "elementwise_max" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseMax())
+      case "elementwise_average" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseAverage())
+      case "elementwise_std_deviation" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.ElementwiseStdDev())
+      case "concatenate" =>
+        AttributeWithAggregator(v, graph_operations.Aggregator.Concatenate[Double]())
```
xandrew, line 81: Would it be hard to also support string vectors for this one? Or we don't support those anyways in other places?

app/com/lynxanalytics/biggraph/graph_operations/Aggregate.scala
```diff
@@ -271,6 +271,29 @@ trait CompoundDoubleAggregator[From]
...
 
+// ElementwiseAggregator executes a given Double aggregator on Vector[Double] attributes,
+// padding them to equal lengths with the configured zero element if needed.
+class Elementwise[Intermediate](val agg: ItemAggregator[Double, Intermediate, Double])
+  extends ItemAggregator[Vector[Double], Vector[Intermediate], Vector[Double]] with Serializable {
+  def zeroElement: Double = 0
+  def zeroIntermediate: Intermediate = agg.zero
+  def zero = Vector[Intermediate]()
+  def merge(a: Vector[Intermediate], b: Vector[Double]) =
+    a.padTo(b.size, zeroIntermediate).zip(b.padTo(a.size, zeroElement)).map {
```
xandrew, line 282: Why do padding? I don't really see a good usecase where this is required, I mean when I naturally end up with different size vectors and I want to pad them with 0s. But I do see that this behavior can hide errors, i.e. I think the vectors should be equal size but they are not due to some upstream problem. I think in this case it's much better to get an error. If there is a really good usecase for this then I'd rather add a separate padding operation.

app/com/lynxanalytics/biggraph/graph_operations/Aggregate.scala
```diff
@@ -271,6 +271,29 @@ trait CompoundDoubleAggregator[From]
   def outputTypeTag(inputTypeTag: TypeTag[From]) = typeTag[Double]
 }
 
+// ElementwiseAggregator executes a given Double aggregator on Vector[Double] attributes,
+// padding them to equal lengths with the configured zero element if needed.
+class Elementwise[Intermediate](val agg: ItemAggregator[Double, Intermediate, Double])
```
xandrew, line 276: Pretty!

app/com/lynxanalytics/biggraph/controllers/FEAggregations.scala
darabos: > Would it be hard to also support string vectors for this one? Or we don't support those anyways in other places?

I tried to make it work for any Vector, but couldn't figure out the type stuff. And by that point I was also of the opinion that this was a dumb idea to begin with. Why would you want to concatenate vectors in random order?

But if you see a good use case, I can have another go at it.

app/com/lynxanalytics/biggraph/graph_operations/Aggregate.scala
darabos: > Why do padding? I don't really see a good usecase where this is required, I mean when I naturally end up with different size vectors and I want to pad them with 0s. But I do see that this behavior can hide errors, i.e. I think the vectors should be equal size but they are not due to some upstream problem. I think in this case it's much better to get an error. If there is a really good usecase for this then I'd rather add a separate padding operation.

I'm sorry, it's because of the zero value. At that point I don't know how wide the vectors are, so I can't create a proper zero vector. I create an empty vector instead and later pad it.

I guess an alternative would be to use `null` as the zero vector and special case it here and there. Would you prefer that?

app/com/lynxanalytics/biggraph/graph_operations/Aggregate.scala
darabos: > Pretty!

I was so surprised this worked out like it did. 😄

app/com/lynxanalytics/biggraph/controllers/FEAggregations.scala
xandrew: I guess it's a poor man's union. As long as Vector is a poor man's set with multiplicities. :) But yeah, I wouldn't get into a jungle fight with types for this. But maybe if easy just add it for strings?

But all this reminds me: we don't even have union and intersection aggregations for sets. Maybe we should?

Alright, based on how hard you see this, either add for strings vector concat and set union and set intersection, or do nothing for now and create an issue?

app/com/lynxanalytics/biggraph/graph_operations/Aggregate.scala
xandrew: Yeah, I think that would be nicer. If you define sg like:
```
def fillIfNull(vector, size): Vector[Intermediate] {
  if (vector == null) ... create zero vector ...
  else ... assert size match and return vector ...
}
```
Then the code will even be simpler.

The only question is finalize. :) I guess you can just assert not null there.

app/com/lynxanalytics/biggraph/graph_operations/Aggregate.scala
xandrew: Scala rocks.

app/com/lynxanalytics/biggraph/graph_operations/Aggregate.scala
darabos: > Yeah, I think that would be nicer. If you define sg like:
> 
> ```
> def fillIfNull(vector, size): Vector[Intermediate] {
>   if (vector == null) ... create zero vector ...
>   else ... assert size match and return vector ...
> }
> ```

I've included the whole other vector as a parameter for better assertion messages.

> Then the code will even be simpler.

I may have made a mistake because it's not simpler. 😓 

> The only question is finalize. :) I guess you can just assert not null there.

In practice `finalize` should never see a `null`. If the set over which it's aggregating is empty, the result is always undefined. (Which is a bit weird. E.g. we may expect `count` to be 0.) So I didn't add an assertion. There will be an exception anyway if this comes up.

app/com/lynxanalytics/biggraph/controllers/FEAggregations.scala
darabos: > Alright, based on how hard you see this, either add for strings vector concat and set union and set intersection, or do nothing for now and create an issue?

Created #81. Sorry.

PR #83, opened 2020-08-19
Changes for macOS build

darabos:
I ended up building LynxKite on macOS somehow, and had to make these two small changes.


xandrew-lynx:
LGTM

Not bad that this was all that needed to change!

PR #84, opened 2020-08-19
Remove KITE_INSTANCE configuration

darabos:
I was editing a `kiterc` and wondered what this was for. It was only used by the log upload, but we haven't used that in years.


xandrew:
LGTM

PR #85, opened 2020-08-28
Fix edge coloring with long fractional numbers

darabos:
Reported by Péter Vakhal. (Thanks!) The edge weights in the example graph are integers, so there's no problem there. The `double` and `string` values in DynamicValue are the same. But with non-integers the two can be different and then the lookup in the colormap does not find the entry.


xandrew:
Thanks, LGTM!

PR #90, opened 2020-09-12
New Neo4j import

darabos:
The existing import can only pull in a single node type or single relationship type. As a table. This makes it hard to get an actual graph from Neo4j to LynxKite.

This PR replaces the operation with a more complex one that can take a whole heterogenous graph.

image

The LynxKite graph will have all the attributes from all the nodes/edges. Each will only be defined for the right types.

image

There is a risk of type collision here: you could have "age" as a string on one node type and as a number on another node type in Neo4j. This will fail to import. I think that's fine.

You configure what to import through a Cypher query:

image

This default query fetches all the edges and all their source and destination nodes. Unfortunately it does not include the isolated nodes. I couldn't figure out a Cypher query to get absolutely everything. I guess Neo4j users are used to this limitation. Or they know how to query the whole graph. 😅


app/com/lynxanalytics/biggraph/controllers/Operation.scala
```diff
@@ -534,10 +531,11 @@ abstract class TableOutputOperation(context: Operation.Context) extends SmartOpe
   override def apply(): Unit = ???
 }
 
-abstract class ImportOperation(context: Operation.Context) extends TableOutputOperation(context) {
-  import MetaGraphManager.StringAsUUID
-  protected def tableFromGuid(guid: String): Table = manager.table(guid.asUUID)
+case class ImportResult(guid: String, parameterSettings: String)
```
xandrew-lynx, line 534: This isn't necessarily a guid anymore. Can we rename? Or that would screw up backward compatibilty? If not possible, add a comment here.

xandrew-lynx (commented):
So far so good! 👍 

I will have to look more at the details, I like the main structure of how you improved the import scaffolding to support projects.

On the query stuff. One thing to consider is to allow two cypher queries, one for edges and one for vertices. Users can leave empty either, which means "all". And the final graph would be taken by combining the vertices from one query with the edger from the other - dropping edges with missing endpoint of course.

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
```diff
@@ -151,27 +145,113 @@ class ImportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+      val project = new RootProjectEditor(CommonProjectState.emptyState)
+      val df = spark.read.format("org.neo4j.spark.DataSource")
+        .option("authentication.type", "basic")
+        .option("authentication.basic.username", params("username"))
+        .option("authentication.basic.password", params("password"))
+        .option("url", params("url"))
+        .option("schema.strategy", "string")
+        .option("query", params("query"))
+        .load
+        .cache
```
xandrew-lynx, line 184: Cache? I have very little faith in spark's cache, I think we are trying not to us it in LK code. Maybe this can also go away if we don't try to get vertex attributes from the edge query and then we can get away with single reads of the neo4j results?

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
```diff
@@ -151,27 +145,113 @@ class ImportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+
+      def getBy(columns: Seq[String], key: String, id: String = null) = {
```
xandrew-lynx, line 186: Huh, this is pretty involved... I'm kinda hoping this might go away with the two query setup, but if not, then it is worth a few lines of explanation.

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
```diff
@@ -151,27 +145,113 @@ class ImportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+      }
+      val oldAttrs = project.edgeAttributes.toMap
+      project.edgeBundle = ebUnion
+      project.edgeAttributes = unifyAttributes(
+        oldAttrs.map {
+          case (name, attr) => name -> attr.pullVia(idUnion.injections(0).reverse)
+        },
+        edges.vertexAttributes.map {
+          case (name, attr) => name -> attr.pullVia(idUnion.injections(1).reverse.concat(imp.embedding))
+        })
```
xandrew-lynx, line 254: Can't we do the unions on the df level and convert to edges only once? I'm worried this is very slow. Also pretty complex.

app/com/lynxanalytics/biggraph/frontend_operations/Operations.scala
```diff
@@ -415,6 +415,100 @@ abstract class ProjectOperations(env: SparkFreeEnvironment) extends OperationReg
...
+      val a1 = project.vertexAttributes(name)
+      val a2 = other.vertexAttributes(name)
+      assert(
+        a1.typeTag.tpe =:= a2.typeTag.tpe,
+        s"Attribute '$name' has conflicting types in the two projects: " +
+          s"(${a1.typeTag.tpe} and ${a2.typeTag.tpe})")
+    }
+  }
+
+  def mergeInto(project: ProjectEditor, other: ProjectViewer): Unit = {
```
xandrew-lynx, line 432: Is this just code moving, or anything to see here?

xandrew-lynx (commented):
Took a closer look, some more detailed comments inline!

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
darabos: > Can't we do the unions on the df level and convert to edges only once? I'm worried this is very slow. Also pretty complex.

Both with edges and nodes the issue is that different types (labels) can have different sets of attributes. So `spark.read.json()` fails on the whole dataset. To make it work I split by label first, run `spark.read.json()` on each type and then union them back, unifying the attributes here.

But now that I think about it I'm not sure how the JSON -> DF parser deals with sparsely defined fields. Is that any different from the situation with different types of nodes? If it just guesses the schema based on the first 10 rows, that is not great. Maybe this can be controlled? I'll take another look.

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
darabos: > Cache? I have very little faith in spark's cache, I think we are trying not to us it in LK code.

We call `persist` all over the place. I think it's pretty important here: we don't want to run the query on Neo4j more than once. And we can't consume it in one go because we have no idea about the schema until we look at the data.

I'll change to `persist(MEMORY_AND_DISK)`.

app/com/lynxanalytics/biggraph/controllers/Operation.scala
darabos: > This isn't necessarily a guid anymore. Can we rename? Or that would screw up backward compatibilty? If not possible, add a comment here.

Renamed to `key`. (It's the key that identifies the imported data somehow.)

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
```diff
@@ -151,27 +145,113 @@ class ImportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
-        params("limit"),
-        params("num_partitions").toInt)
+    override def getOutputs(): Map[BoxOutput, BoxOutputState] = {
+      params.validate()
+      assert(params("imported_table").nonEmpty, "You have to import the data first.")
+      assert(!areSettingsStale, areSettingsStaleReplyMessage)
+      import CommonProjectState._
+      val state = json.Json.parse(params("imported_table")).as[CommonProjectState]
+      val project = new RootProjectEditor(state)
+      Map(context.box.output(context.meta.outputs(0)) -> BoxOutputState.from(project))
```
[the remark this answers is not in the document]
darabos: > Why do you need the editor here? There is a `BoxOutputState.from` using `CommonProjectState`.

Nice, thanks.

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
darabos: > Huh, this is pretty involved... I'm kinda hoping this might go away with the two query setup, but if not, then it is worth a few lines of explanation.

It's independent of whether we do one or two queries. But it's gone now.

The Neo4j connector was failing on schema detection if I had more than one kind of label in the result. So I thought I had to split by label before parsing the JSON. But no. It works fine if I don't split. It's also much faster. 😓

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
darabos: > But now that I think about it I'm not sure how the JSON -> DF parser deals with sparsely defined fields. Is that any different from the situation with different types of nodes? If it just guesses the schema based on the first 10 rows, that is not great. Maybe this can be controlled? I'll take another look.

It reads the whole input to determine the schema. (https://github.com/apache/spark/blob/v2.4.7/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/json/JsonInferSchema.scala) It works fine with the whole heterogenous node set from Neo4j.

app/com/lynxanalytics/biggraph/frontend_operations/Operations.scala
darabos: > Is this just code moving, or anything to see here?

I can actually move it back now because I ended up not using it. Should I? Or do you like it here?

darabos:
PTAL
image

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
```diff
@@ -151,27 +145,85 @@ class ImportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+        if (params("vertex_query").isEmpty) {
+          // Create vertices from the edge endpoints.
+          val vdf = df.select("`<source.id>`")
+            .union(df.select("`<target.id>`"))
+            .withColumnRenamed("<source.id>", "<id>")
+            .distinct
+          val table = graph_operations.ImportDataFrame.run(vdf)
+          val vs = table.toProject
+          project.vertexSet = vs.vertexSet
+          project.vertexAttributes = vs.viewer.vertexAttributes
```
xandrew-lynx, line 210: Why do we need to do this? Why not just read the edges and apply `Use table as graph` (well, more precisely `graph_operations.VerticesToEdges`)?

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
```diff
@@ -151,27 +145,85 @@ class ImportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+        import spark.implicits._
+        val json = spark.read.format("org.neo4j.spark.DataSource")
+          .option("authentication.type", "basic")
+          .option("authentication.basic.username", params("username"))
+          .option("authentication.basic.password", params("password"))
+          .option("url", params("url"))
+          .option("schema.strategy", "string")
+          .option("query", query)
+          .load
+          .persist(org.apache.spark.storage.StorageLevel.MEMORY_AND_DISK)
```
xandrew-lynx, line 186: I'm still not convinced we need this. Now (if we can switch to using `VerticesToEdges` below) we do seem to get away with a single scan through the results. Yes, spark need to do something more to figure out the schema, but I don't think we should be trying to support that on our end by caching, it should really be spark's responsibility to do something smart about that. And even if it doesn't do anything smart, this is too late to help him here: by the time this persist is executed the schema is figured out, no?

So at the end of the day I believe that now we add all the data to memory/disk when we read through it and then we never read it again.

app/com/lynxanalytics/biggraph/frontend_operations/Operations.scala
xandrew-lynx: I don't really have a strong preference either way. I think it's fine here, but also fine to leave it as it was.

web/app/help/operations/import-from-neo4j.asciidoc
```diff
@@ -0,0 +1,29 @@
...
+Import a graph from the https://neo4j.com/[Neo4j] graph database.
+
+Neo4j does not have a strict schema. Different nodes may have different attributes.
+In LynxKite the list of vertex attributes is defined for the whole graph.
+But each vertex may leave any attribute undefined.
+
+If you import Neo4j nodes that have different attributes, such as movies that have
+a `title` and actors that have a `name`, the resulting graph will have both `title` and
+`name` attributes. `title` will only be defined on movies, `name` will only be defined
+on actors.
```
xandrew-lynx, line 12: I think we may need to explain that there can be problems if the same attribute has different type on different kinds. And that the way to fix that is to make the query more specific somehow and not import (both) of the conflicting attributes.

xandrew-lynx (approved):
Beautiful!

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
darabos: > I'm still not convinced we need this. Now (if we can switch to using `VerticesToEdges` below) we do seem to get away with a single scan through the results. Yes, spark need to do something more to figure out the schema, but I don't think we should be trying to support that on our end by caching, it should really be spark's responsibility to do something smart about that.

Spark doesn't take responsibility for that.

```
scala> val r = Seq("{\"x\": 1}").toDF.rdd.map(x => { println("quack"); x.getString(0) })                                                                               
r: org.apache.spark.rdd.RDD[String] = MapPartitionsRDD[31] at map at <console>:23

scala> spark.read.json(r).show
quack
quack
+---+
|  x|
+---+
|  1|
+---+
```

One quack for figuring out the schema, another quack for getting the data.

> And even if it doesn't do anything smart, this is too late to help him here: by the time this persist is executed the schema is figured out, no?

No. This doesn't do anything yet. `spark.read.json()` (two lines below) figures out the schema. Then the LynxKite operations do the read at some point down the line.

It's not even just a performance question. The data may change in Neo4j between our two reads. (I guess this is an argument against the two separate queries too, but let's not worry about that now.)

app/com/lynxanalytics/biggraph/frontend_operations/ImportOperations.scala
darabos: > Why do we need to do this? Why not just read the edges and apply `Use table as graph` (well, more precisely `graph_operations.VerticesToEdges`)?

Done. It's not less code. Unfortunately we have to duplicate some code. The two operation outputs contain basically the same things (edges, embedding) but they are unrelated types, so we can't handle them together.

But still it should be faster. Thanks!

app/com/lynxanalytics/biggraph/controllers/Operation.scala
darabos: > Renamed to `key`. (It's the key that identifies the imported data somehow.)

Sorry, I named it back. The name is used in the Python API too and I don't want trouble.

PR #91, opened 2020-09-30
Neo4j export

darabos:
This is part 1: exporting attributes for existing nodes.

There's an option to set `node.keys` and let them build the query. But if I use that, the label is a must. If I write the same query manually, I can leave it off. (http://5.9.211.195:8000/neo4j-spark-docs/1.0.0/writing.html#bookmark-write-node)

Open tasks:
- [x] Make sure this works if the keys are not defined everywhere.
- [x] Attribute export for edges.
- [ ] Edge export for existing nodes. (I don't think this is important.)
- [x] Export whole graph as new stuff.
- [x] Documentation.
- [x] Tests. (Maybe when the final Neo4j Spark Connector is released.)

### DIFF
--- a/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
+++ b/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
@@ -172,11 +172,11 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
       Map(context.box.output(
         context.meta.outputs(0)) -> BoxOutputState.from(exportResult, params.toMap - "password"))
     }
+    def getAttribute(a: String): com.lynxanalytics.biggraph.graph_api.Attribute[_]
     def exportResult() = {
       val keys = splitParam("keys")
       val attrs = (splitParam("to_export") ++ keys).toSet.toList
-      val t = graph_operations.AttributesToTable.run(
-        attrs.map(a => a -> project.vertexAttributes(a)))
+      val t = graph_operations.AttributesToTable.run(attrs.map(a => a -> getAttribute(a)))
       assert(keys.nonEmpty, "You have to choose one or more attributes to use as the keys for identifying the nodes in Neo4j.")
       val op = graph_operations.ExportAttributesToNeo4j(
         params("url"), params("username"), params("password"), params("labels"), keys,
@@ -196,6 +196,7 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
           options = project.vertexAttrList.filter(!_.id.startsWith("<")), multipleChoice = true),
         Choice("to_export", "Exported attributes", options = project.vertexAttrList, multipleChoice = true))
       val nodesOrRelationships = "nodes"
+      def getAttribute(a: String) = project.vertexAttributes(a)
     })
   registerOp(
     "Export edge attributes to Neo4j", defaultIcon, ExportOperations,
@@ -207,7 +208,8 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
           // Cannot join on internal ID ("<id>") and stuff like that.
           options = project.edgeAttrList.filter(!_.id.startsWith("<")), multipleChoice = true),
         Choice("to_export", "Exported attributes", options = project.edgeAttrList, multipleChoice = true))
-      val nodesOrRelationships = "edges"
+      val nodesOrRelationships = "relationships"
+      def getAttribute(a: String) = project.edgeAttributes(a)
     })
 
   registerOp(
--- a/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
+++ b/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
@@ -190,7 +190,9 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
       import Operation.Implicits._
       params ++= List(
         Param("labels", "Node labels", defaultValue = ""),
-        Choice("keys", "Attribute to use as key", options = project.vertexAttrList, multipleChoice = true),
+        Choice("keys", "Attribute to use as key",
+          // Cannot join on internal ID ("<id>") and stuff like that.
+          options = project.vertexAttrList.filter(!_.id.startsWith("<")), multipleChoice = true),
         Choice("to_export", "Exported attributes", options = project.vertexAttrList, multipleChoice = true))
       val nodesOrRelationships = "nodes"
     })
@@ -200,7 +202,9 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
       import Operation.Implicits._
       params ++= List(
         Param("labels", "Relationship labels", defaultValue = ""),
-        Choice("keys", "Attribute to use as key", options = project.edgeAttrList, multipleChoice = true),
+        Choice("keys", "Attribute to use as key",
+          // Cannot join on internal ID ("<id>") and stuff like that.
+          options = project.edgeAttrList.filter(!_.id.startsWith("<")), multipleChoice = true),
         Choice("to_export", "Exported attributes", options = project.edgeAttrList, multipleChoice = true))
       val nodesOrRelationships = "edges"
     })
--- a/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
+++ b/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
@@ -164,7 +164,7 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
     lazy val project = projectInput("graph")
     val nodesOrRelationships: String
     override def enabled = FEStatus.enabled
-    override def trigger(wc: WorkspaceController, gdc: GraphDrawingController) = {
+    override def trigger(wc: WorkspaceController, gdc: GraphDrawingController): scala.concurrent.Future[Unit] = {
       gdc.getComputeBoxResult(List(exportResult.gUID))
     }
     override def getOutputs(): Map[BoxOutput, BoxOutputState] = {
@@ -225,7 +225,7 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
           options = List(FEOption("", "")) ++ project.edgeAttrList[String]))
       lazy val project = projectInput("graph")
       override def enabled = FEStatus.enabled
-      override def trigger(wc: WorkspaceController, gdc: GraphDrawingController) = {
+      override def trigger(wc: WorkspaceController, gdc: GraphDrawingController): scala.concurrent.Future[Unit] = {
         gdc.getComputeBoxResult(List(exportResult.gUID))
       }
       override def getOutputs(): Map[BoxOutput, BoxOutputState] = {
--- a/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
+++ b/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
@@ -160,7 +160,7 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
       Param("url", "Neo4j connection", defaultValue = "bolt://localhost:7687"),
       Param("username", "Neo4j username", defaultValue = "neo4j"),
       Param("password", "Neo4j password", defaultValue = "neo4j"),
-      NonNegInt("version", "Snapshot revision ID", default = 1))
+      NonNegInt("version", "Export repetition ID", default = 1))
     lazy val project = projectInput("graph")
     val nodesOrRelationships: String
     override def enabled = FEStatus.enabled
@@ -191,7 +191,7 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
       import Operation.Implicits._
       params ++= List(
         Param("labels", "Node labels", defaultValue = ""),
-        Choice("keys", "Attribute to use as key",
+        Choice("keys", "Attribute(s) to use as key",
           // Cannot join on internal ID ("<id>") and stuff like that.
           options = project.vertexAttrList.filter(!_.id.startsWith("<")), multipleChoice = true),
         Choice("to_export", "Exported attributes", options = project.vertexAttrList, multipleChoice = true))
@@ -203,7 +203,7 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
       import Operation.Implicits._
       params ++= List(
         Param("labels", "Relationship labels", defaultValue = ""),
-        Choice("keys", "Attribute to use as key",
+        Choice("keys", "Attribute(s) to use as key",
           // Cannot join on internal ID ("<id>") and stuff like that.
           options = project.edgeAttrList.filter(!_.id.startsWith("<")), multipleChoice = true),
         Choice("to_export", "Exported attributes", options = project.edgeAttrList, multipleChoice = true))
@@ -218,7 +218,7 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
         Param("url", "Neo4j connection", defaultValue = "bolt://localhost:7687"),
         Param("username", "Neo4j username", defaultValue = "neo4j"),
         Param("password", "Neo4j password", defaultValue = "neo4j"),
-        NonNegInt("version", "Snapshot revision ID", default = 1),
+        NonNegInt("version", "Export repetition ID", default = 1),
         Choice("node_labels", "Attribute with node labels",
           options = List(FEOption("", "")) ++ project.vertexAttrList[String]),
         Choice("relationship_type", "Attribute with relationship type",
--- a/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
+++ b/app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
@@ -2,6 +2,7 @@ package com.lynxanalytics.biggraph.frontend_operations
 
 import com.lynxanalytics.biggraph.SparkFreeEnvironment
 import com.lynxanalytics.biggraph.graph_api.Scripting._
+import com.lynxanalytics.biggraph.graph_util.Scripting._
 import com.lynxanalytics.biggraph.graph_operations
 import com.lynxanalytics.biggraph.controllers._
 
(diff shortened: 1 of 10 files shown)


app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
```diff
@@ -0,0 +1,53 @@
...
+  def execute(
+    inputDatas: DataSet,
+    o: ExportAttributesToNeo4j.Output,
+    output: OutputBuilder,
+    rc: RuntimeContext): Unit = {
+    implicit val ds = inputDatas
+    implicit val sd = rc.sparkDomain
+    val df = inputs.t.df
+    val keyMatch = keys.map(k => s"`$k`: event.`$k`").mkString(", ")
+    val query = s"MERGE (n$labels {$keyMatch}) SET n += event"
```
xandrew-lynx, line 40: Is `event` a cypher reserved word here? It refers to the actual record streaming in?

xandrew-lynx (commented):
:+1: 

So if key is not actually a key, then what happens? We will set the parameter for all nodes that match on they "keys", right? I'm not necessarily saying this is bad, just making sure I understand what's going on.

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
darabos: > Is `event` a cypher reserved word here? It refers to the actual record streaming in?

Yes, it's the actual record. The connector adds `UNWIND $events AS event` before the query we specify. I'm not sure what is reserved, but that's a part outside of our control at least.

darabos:
> So if key is not actually a key, then what happens? We will set the parameter for all nodes that match on they "keys", right? I'm not necessarily saying this is bad, just making sure I understand what's going on.

I think so. I'll be sure to note this in the docs. I think it's just how Neo4j updates work. (Much like SQL updates.)

darabos:
> So if key is not actually a key, then what happens? We will set the parameter for all nodes that match on they "keys", right? I'm not necessarily saying this is bad, just making sure I understand what's going on.

The Neo4j import creates an `<id>` field, which would make an obvious key. But since the ID is not a property, it does not work. I've checked how it can be made to work — it cannot. As https://stackoverflow.com/questions/27729489/neo4j-merge-statement-with-id and the linked answer you must have your hand-crafted keys in Neo4j.

darabos:
Screenshot 2020-10-01 at 15 23 32

Isn't that a joy to see? 😄   I got this with gender and comment selected as node label and relationship type. If you omit one or both:

Screenshot 2020-10-01 at 15 21 40
Screenshot 2020-10-01 at 15 21 01

PTAL

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
```diff
@@ -0,0 +1,135 @@
...
+      .withColumn(DSTID, F.concat(guid, F.col(SRCDST + "._2")))
+      .drop(SRCDST)
+    if (nodeLabelsColumn.isEmpty) {
+      neo.send(vs, s"""
+        CREATE (n)
+        SET n += event
+      """)
+    } else {
+      neo.send(vs, s"""
+        CALL apoc.create.node(split(event.`$nodeLabelsColumn`, ','), event) YIELD node
```
darabos, line 115: See https://neo4j.com/labs/apoc/4.1/overview/apoc.create/apoc.create.node/.
This is the way to create a node with a dynamically picked label. Requires `apoc` to be installed.

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
```diff
@@ -0,0 +1,135 @@
...
+    if (relationshipTypeColumn.isEmpty) {
+      neo.send(es, s"""
+      MATCH (src {`$VID`: event.`$SRCID`}), (dst {`$VID`: event.`$DSTID`})
+      CREATE (src)-[r:EDGE]->(dst)
+      SET r += event
+    """)
+    } else {
+      neo.send(es, s"""
+      MATCH (src {`$VID`: event.`$SRCID`}), (dst {`$VID`: event.`$DSTID`})
+      CALL apoc.create.relationship(src, event.`$relationshipTypeColumn`, event, dst) YIELD rel
```
darabos, line 128: See https://neo4j.com/labs/apoc/4.1/overview/apoc.create/apoc.create.relationship/.
Same deal.

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
```diff
@@ -0,0 +1,135 @@
...
+    "version" -> version)
+  def execute(
+    inputDatas: DataSet,
+    o: ExportGraphToNeo4j.Output,
+    output: OutputBuilder,
+    rc: RuntimeContext): Unit = {
+    implicit val ds = inputDatas
+    val F = spark.sql.functions
+    // Prefix the internal IDs with this operation's GUID so different exports don't collide.
+    val guid = F.lit(this.gUID.toString + "-")
```
darabos, line 101: Does this make sense?

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
```diff
@@ -153,4 +154,96 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+  abstract class Neo4jAttributeExport(context: Operation.Context) extends TriggerableOperation(context) {
+    params ++= List(
+      Param("url", "Neo4j connection", defaultValue = "bolt://localhost:7687"),
+      Param("username", "Neo4j username", defaultValue = "neo4j"),
+      Param("password", "Neo4j password", defaultValue = "neo4j"),
+      NonNegInt("version", "Snapshot revision ID", default = 1))
+    lazy val project = projectInput("graph")
+    val nodesOrRelationships: String
+    override def enabled = FEStatus.enabled
+    override def trigger(wc: WorkspaceController, gdc: GraphDrawingController) = {
```
xandrew-lynx, line 167:
```suggestion
    override def trigger(wc: WorkspaceController, gdc: GraphDrawingController): scala.concurrent.Future[Unit] = {
```

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
```diff
@@ -153,4 +154,96 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+    }
+  }
+
+  registerOp(
+    "Export vertex attributes to Neo4j", defaultIcon, ExportOperations,
+    List("graph"), List("exported"), new Neo4jAttributeExport(_) {
+      import Operation.Implicits._
+      params ++= List(
+        Param("labels", "Node labels", defaultValue = ""),
+        Choice("keys", "Attribute to use as key",
```
xandrew-lynx, line 194:
```suggestion
        Choice("keys", "Attributes to use as key",
```

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
```diff
@@ -153,4 +154,96 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+        Choice("to_export", "Exported attributes", options = project.vertexAttrList, multipleChoice = true))
+      val nodesOrRelationships = "nodes"
+    })
+  registerOp(
+    "Export edge attributes to Neo4j", defaultIcon, ExportOperations,
+    List("graph"), List("exported"), new Neo4jAttributeExport(_) {
+      import Operation.Implicits._
+      params ++= List(
+        Param("labels", "Relationship labels", defaultValue = ""),
+        Choice("keys", "Attribute to use as key",
```
xandrew-lynx, line 206:
```suggestion
        Choice("keys", "Attributes to use as key",
```

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
```diff
@@ -153,4 +154,96 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+        Param("username", "Neo4j username", defaultValue = "neo4j"),
+        Param("password", "Neo4j password", defaultValue = "neo4j"),
+        NonNegInt("version", "Snapshot revision ID", default = 1),
+        Choice("node_labels", "Attribute with node labels",
+          options = List(FEOption("", "")) ++ project.vertexAttrList[String]),
+        Choice("relationship_type", "Attribute with relationship type",
+          options = List(FEOption("", "")) ++ project.edgeAttrList[String]))
+      lazy val project = projectInput("graph")
+      override def enabled = FEStatus.enabled
+      override def trigger(wc: WorkspaceController, gdc: GraphDrawingController) = {
```
xandrew-lynx, line 228:
```suggestion
      override def trigger(wc: WorkspaceController, gdc: GraphDrawingController): scala.concurrent.Future[Unit] = {
```

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
```diff
@@ -0,0 +1,135 @@
...
+      .option("query", query)
+      .save()
+  }
+}
+
+case class ExportAttributesToNeo4j(
+    url: String, username: String, password: String, labels: String, keys: Seq[String],
+    version: Long, nodesOrRelationships: String)
+  extends SparkOperation[ExportAttributesToNeo4j.Input, ExportAttributesToNeo4j.Output] {
+  val neo = Neo4jConnection(url, username, password)
```
xandrew-lynx, line 38: When I first saw this (and of course didn't realize I just reviewed the definition of this class :( ) I thought why would you put a connection kind of class up here, why not create the connection only when needed, in execute. But of course, there is no connection being made here, it just collects some metadata which will be used in a helper method of this object. So then what about naming it differently? E.g. Neo4JHelper?

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
xandrew-lynx: Yeah, I think this is perfect! But what about adding GUID as vertex and edge attribute as well? That would allow users to clean up results of somehow screwed up exports with relative ease. Maybe we can add export date as well?

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
```diff
@@ -153,4 +154,96 @@ class ExportOperations(env: SparkFreeEnvironment) extends OperationRegistry {
...
+  registerOp(
+    "Export graph to Neo4j", defaultIcon, ExportOperations,
+    List("graph"), List("exported"), new TriggerableOperation(_) {
+      import Operation.Implicits._
+      params ++= List(
+        Param("url", "Neo4j connection", defaultValue = "bolt://localhost:7687"),
+        Param("username", "Neo4j username", defaultValue = "neo4j"),
+        Param("password", "Neo4j password", defaultValue = "neo4j"),
+        NonNegInt("version", "Snapshot revision ID", default = 1),
+        Choice("node_labels", "Attribute with node labels",
```
xandrew-lynx, line 222: Mention that the value of this attribute can be a comma separated list of label values if user wants multiple labels. Would be more elegant to allow vectors here, but probably now is not the time.

xandrew-lynx (commented):
This looks great, thanks a lot!

I only added some nits.

But I wouldn't delay the testing until after the release. If things go well, this will be a very high profile feature. I understand that this is substantial work because we somehow have to get a testing neo4j instance up, so this will probably remove our chances for a release today, but still. Or is this going to be simpler after the final connector is released somehow?

xandrew-lynx:
On the screenshot: indeed feels cool! :) And if you just export twice, you will simply see two copies, right?

darabos:
> But I wouldn't delay the testing until after the release. If things go well, this will be a very high profile feature. I understand that this is substantial work because we somehow have to get a testing neo4j instance up, so this will probably remove our chances for a release today, but still. Or is this going to be simpler after the final connector is released somehow?

There's a whole Neo4j test library which we were using for the old import. Instead of starting a full Neo4j instance I think it sets something up in-process. Now we can't use it because it's an old version. I'll look around to see if there is a replacement.

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
darabos: I've intentionally put it in the singular. I think 99% of the cases you have a single key. You need to somehow identify nodes, right? If you use first name + last name + birth date you will have issues.

But it's fairly clearly a multiple-choice parameter and it's documented in the help. (Not yet committed.)

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
darabos: > Yeah, I think this is perfect! But what about adding GUID as vertex and edge attribute as well? That would allow users to clean up results of somehow screwed up exports with relative ease. Maybe we can add export date as well?

Actually the timestamp is better than the GUID. Maybe you've recreated your LynxKite instance. Maybe two LynxKite instances use the same Neo4j instance. If everyone exports the example graph, you can end up with two exports with the same GUID quick enough.

I'll switch to just using the timestamp. Easier to interpret too.

xandrew-lynx: 👍

xandrew-lynx:
On the testing: yes, please look around. Of course, if the conclusion is that now it's hard/ugly but in the not-too-far future we a see good solution on the horizon then sure, we can wait of it.

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
xandrew-lynx: Hmm, the fact that the name and the kind conflicts is exactly what I would find strange/confusing and I'm not sure this is a good way to encourage people to use a single key attribute.  What about attribute(s) as a compromise? :)

darabos: > What about attribute(s) as a compromise? :)

Done.

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
darabos: Done.

app/com/lynxanalytics/biggraph/frontend_operations/ExportOperations.scala
darabos: > Mention that the value of this attribute can be a comma separated list of label values if user wants multiple labels. Would be more elegant to allow vectors here, but probably now is not the time.

I expect you usually just use one label. Described the commas in the help.

app/com/lynxanalytics/biggraph/graph_operations/ExportToNeo4j.scala
darabos: > When I first saw this (and of course didn't realize I just reviewed the definition of this class :( ) I thought why would you put a connection kind of class up here, why not create the connection only when needed, in execute. But of course, there is no connection being made here, it just collects some metadata which will be used in a helper method of this object. So then what about naming it differently? E.g. Neo4JHelper?

I changed it to Neo4jConnectionParameters.

darabos (commented):
Thanks for the comments! You can review the docs now and I'll try to do the tests.

web/app/help/operations/export-edge-attributes-to-neo4j.asciidoc
```diff
@@ -0,0 +1,38 @@
+### Export edge attributes to Neo4j
+
+Exports edge attributes from a graph in LynxKite to a
+corresponding graph in https://neo4j.com/[Neo4j].
+
+The relationships in Neo4j are identified by a key property (or properties).
+You must have a corresponding edge attribute in LynxKite by the same name.
+This will be used to find the right relationship to update in Neo4j.
+
```
xandrew-lynx, line 9: I think we need to explicitly spell out what happens if multiple edges match a given key value combination
a) in Neo4j
b) in LynxKite

web/app/help/operations/export-graph-to-neo4j.asciidoc
```diff
@@ -0,0 +1,37 @@
...
+
+[p-version]#Export repetition ID#::
+LynxKite only re-computes outputs if parameters or inputs have changed.
+This is true for exports too. If you want to repeat a previous export, you can increase this
+export repetition ID parameter.
+
+[p-node_labels]#Node labels#::
+A string vertex attribute that is a comma-separated list of labels to apply to the newly
+created nodes. Optional. You must have https://neo4j.com/developer/neo4j-apoc/[Neo4j APOC]
+installed on the Neo4j instance to use this.
```
xandrew-lynx, line 30: If omitted, the nodes will have no label, right? Is this acceptable in Neo4J then?

web/app/help/operations/export-graph-to-neo4j.asciidoc
```diff
@@ -0,0 +1,37 @@
...
+
+[p-relationship_type]#Attribute with relationship type#::
+A string edge attribute that specifies the relationship type for each newly created relationship.
+Optional. You must have https://neo4j.com/developer/neo4j-apoc/[Neo4j APOC]
+installed on the Neo4j instance to use this.
```
xandrew-lynx, line 35: Interesting. So edges can have only one relationship type, but vertices can have multiple labels? Sorry for bothering you with basic neo4j questions..

xandrew-lynx (approved):

xandrew-lynx (approved):
Nice! These tests ended up being very pretty and readable!

web/app/help/operations/export-edge-attributes-to-neo4j.asciidoc
darabos: > I think we need to explicitly spell out what happens if multiple edges match a given key value combination
> a) in Neo4j
> b) in LynxKite

Done. I've also added the full Cypher query to the docs. I imagine this may be of interest to advanced Neo4j users and it's not too scary.

web/app/help/operations/export-graph-to-neo4j.asciidoc
darabos: > If omitted, the nodes will have no label, right? Is this acceptable in Neo4J then?

Surprisingly, yes.

web/app/help/operations/export-graph-to-neo4j.asciidoc
darabos: > Interesting. So edges can have only one relationship type, but vertices can have multiple labels? Sorry for bothering you with basic neo4j questions..

Yes. Edges must always have a type, and they can only have a single type. Nodes can have multiple labels or no labels. It's pretty clear that you didn't design this system. 😄

darabos:
> Nice! These tests ended up being very pretty and readable!

They are actually surprisingly fast too. Around 10-30 seconds. 👍  https://www.testcontainers.org/

PR #86, opened 2020-09-05
Add support for Delta format

bramrodenburg:
**Description**
In this pull request I have added support for reading and writing to Delta Lake in LynxKite. "Delta Lake is an open-source storage layer that brings ACID transactions to to Apache Spark and big data workloads." \[1\] 

I used Delta version `0.6.1`, since versions >= `0.7.x` require Spark 3.0. 

\[1\]: https://docs.delta.io/latest/index.html

PR #99, opened 2020-10-14
Add new import/export operations to Python API

darabos:
Forgot these in #63 , #86, #90, #91. But it's not too late to release a 4.1.0 Python package.


darabos:
It's live at https://pypi.org/project/lynxkite-client/.

xandrew-lynx:
:+1: Thanks a lot!

PR #102, opened 2020-11-12
NetworKit integration

darabos:
Super early state, but I can finally call NetworKit from Go. It's similar to Jano's solution from a year ago, but doesn't require hand-crafted wrappers. SWIG generates them just fine!

For now I only communicate "scalars" between the two systems. Passing arrays was another hurdle in Jano's PR. We will see.

(Internal link for his PR: https://github.com/biggraph/biggraph/pull/8676)


Gabitany:
Hurray and congratulations + best of luck with the arrays..😃

sphynx/networkit/networkit.swigcxx
```diff
@@ -0,0 +1,23 @@
+%module networkit
+
+%{
+/*
+#include "example.h"
+*/
+#include <networkit/centrality/Betweenness.hpp>
+#include <networkit/generators/BarabasiAlbertGenerator.hpp>
+using namespace NetworKit;
+%}
```
xandrew-lynx, line 10: Sorry, I'm a total SWIG noob, but probably it's worth for me to understand a bit what's going on here for the sake of future PRs. So some questions:
- What's the deal with the stuff enclosed in `%{}%`? Why does it "duplicate" in a sense what's included below?
- What's with the apparently commented out `example.h` includes?
- Most importantly, what's the effect of this file? We ask swig to create wrappers for everything that we `%include`? Or that we `#inlcude`? Or something else? :)

darabos: > Sorry, I'm a total SWIG noob

That makes two of us. 😄 

> * What's the deal with the stuff enclosed in `%{}%`? Why does it "duplicate" in a sense what's included below?

A C++ language wrapper will be generated from this file. The stuff in `%{}%` will be put into the C++ code verbatim. It's for the C++ compiler.

The other part defines the interface that SWIG will generate wrappers for. The language for defining the interface is _almost_ C++. So we can `%include` unaltered C++ headers in many cases.

There are just two tricks here: `#define final` because the SWIG parser does not know about the `final` keyword that NetworKit classes use. And `class Graph` because SWIG couldn't parse `Graph.hpp` and I also didn't need to wrap it for this test. I'll sure need it later, but I can write a simpler version to define a limited interface for SWIG to wrap. I don't mind if I have to do that for a few base classes as long as wrapping the operations is cheap.

> * What's with the apparently commented out `example.h` includes?

I forgot to delete them. 😓

xandrew-lynx: Thanks a lot, this helps! One more thing. So what is a C++ wrapper in this context? Here is my theory, tell me whether this is accurate...

So, it is still C++ code, defining C++ stuff (functions? classes?). These stuff use the original C++ stuff for their implementation. The reason we need them, and can't use the original "stuff" is that go is limited in what kind of stuff it can directly call. So we are wrapping non-go-compliant kind of stuff within go-compliant kind of stuff.

Then, there must be one more thing being generated here: go code which implements or declares the calls to the "C++ wrapper", the go-compliant stuff from above.

darabos: > Thanks a lot, this helps!

I'm also adding a few comments to explain this in the file.

> One more thing. So what is a C++ wrapper in this context? Here is my theory, tell me whether this is accurate...
> 
> So, it is still C++ code, defining C++ stuff (functions? classes?). These stuff use the original C++ stuff for their implementation. The reason we need them, and can't use the original "stuff" is that go is limited in what kind of stuff it can directly call. So we are wrapping non-go-compliant kind of stuff within go-compliant kind of stuff.
> 
> Then, there must be one more thing being generated here: go code which implements or declares the calls to the "C++ wrapper", the go-compliant stuff from above.

That's exactly how it seems to work. It generates a bunch of files like `networkit_wrap.cxx` and `_networkit_swig.cgo1.go`. Go cannot straight up access C++, so I think the C++ side of the wrapper creates a C interface. (And does who knows what else. I haven't looked.)

There are two things I like:
- This is all handled by the Go tooling. I just run `go build` and when it sees the `.swigcxx` file it runs a bunch of commands to take care of everything.
- SWIG has no problem automatically wrapping the algorithms. Once the basics are set up I think it will be easy to access the whole library.

I've got the Sphynx &rarr; NetworKit direction working for passing in a graph now. Unfortunately the NetworKit `Graph` object is much more complex than our `EdgeBundle`, so this is far from zero-copy. At least it doesn't go through disk or Python. 😄

darabos:
Looks like it's working!

image

It's time to start thinking about where we want to put the hundreds of new algorithms!

xandrew-lynx (commented):
Thanks, promising!

Is it possible you forgot to add some go files? I don't seem to find where the sphynx operation is actually implemented.

darabos:
> Is it possible you forgot to add some go files? I don't seem to find where the sphynx operation is actually implemented.

Wow, I forgot to add a ton of them. Sorry.

sphynx/lynxkite-sphynx/networkit_compute_attribute.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/lynxkite/lynxkite/sphynx/networkit"
+)
+
+func ToNetworKit(vs *VertexSet, es *EdgeBundle) networkit.Graph {
+	builder := networkit.NewGraphBuilder(uint64(len(vs.MappingToUnordered)))
+	for i := range es.Src {
+		builder.AddHalfEdge(uint64(es.Src[i]), uint64(es.Dst[i]))
+	}
+	return builder.ToGraph(true)
+}
```
xandrew-lynx, line 14: Isn't this already in additions.go?

sphynx/lynxkite-sphynx/networkit_compute_attribute.go
```diff
@@ -0,0 +1,41 @@
...
+			switch ea.GetStringParam("op") {
+			case "betweenness":
+				b := networkit.NewBetweenness(g)
+				b.Run()
+				result = b.Scores()
+			}
+			attr.Values = networkit.ToDoubleSlice(result)
+			for i := range attr.Defined {
+				attr.Defined[i] = true
+			}
```
xandrew-lynx, line 36: Oh, it's been a while since I looked at go code. Really, this is the way you fill a slice with a constant value? *Sigh*

sphynx/networkit/additions.h
```diff
@@ -0,0 +1,11 @@
...
+#include <vector>
+#include <networkit/Globals.hpp>
+#include <networkit/graph/Graph.hpp>
+
+struct EdgeList {
+  NetworKit::count n;
+  std::vector<unsigned long long> src, dst;
+};
+
+EdgeList graphToEdgeList(NetworKit::Graph& g);
```
xandrew-lynx, line 11: Aha, so all this is ours. But why do we need these two copy steps, isn't one enough? Is it not possible to pass over from go a simple pointer to a pre-allocated slice - hopefully looking like a `*unsigned long long` from he C++ side - and fill things up directly into the slice? You may need to first get `n` over to go in a separate call.

xandrew-lynx (commented):
Yeah, things add up more now. :)

sphynx/networkit/additions.go
```diff
@@ -0,0 +1,55 @@
...
+type EdgeBundle struct {
+	Src         []SphynxId
+	Dst         []SphynxId
+	EdgeMapping []int64
+}
+
+type VertexSet struct {
+	MappingToUnordered []int64
+	MappingToOrdered   map[int64]SphynxId
+}
```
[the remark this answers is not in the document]
darabos: > How come we have to duplicate these here? And no need for mutex anymore?

I certainly don't want to leave it here, sorry. This is garbage.

I want to have a function for converting between Sphynx and NetworKit graphs. But where to put it? I first put it in the `networkit` package. But it doesn't have the Sphynx types. So I just repeated them here for the purpose of the test.

But for the real operation I copied it to the `main` package.

So options are:
- It's fine in `main`. Delete it from `networkit` and move the test to `main`.
- We want it in `networkit`. For this we need a third package to define the Sphynx types, so we can import them in both `networkit` and `main`.

sphynx/networkit/additions.go
```diff
@@ -0,0 +1,55 @@
...
+	for i := range s {
+		s[i] = SphynxId(v.Get(i))
+	}
+	return s
+}
+
+func ToNetworKit(vs VertexSet, es EdgeBundle) Graph {
+	builder := NewGraphBuilder(uint64(len(vs.MappingToUnordered)))
+	for i := range es.Src {
+		builder.AddHalfEdge(uint64(es.Src[i]), uint64(es.Dst[i]))
```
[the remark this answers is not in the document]
darabos: > Half? What about the other half? :) That's somehow handled by `ToGraph` below I suppose?
> 
> And wow, this is indeed not no-copy. Wasn't we originally motivated by NetworKit data structures when designing Sphynx so that we don't need this?

Yeah I had the same shock. I looked a lot at NetworKit code, e.g. how their file readers create graphs. It's always extremely complex and uses unsafe APIs. ([NetworkitBinaryReader.cpp](https://github.com/networkit/networkit/blob/7.1/networkit/cpp/io/NetworkitBinaryReader.cpp)) Jano's PR used this GraphBuilder interface too and it's [documented](https://github.com/networkit/networkit/blob/7.1/include/networkit/graph/GraphBuilder.hpp) as the way to "speed up" building a graph.

> Or maybe that was only for attributes with this ordered/unordered thing?

Ah, I  think that's it! For attributes it's fairly straightforward.

> What is their internal represenation? Neighbor lists for both incoming and outgoing edges? Just guessing based on the half edge notation...

Yes.

```c++
std::vector<std::vector<node>> inEdges;
std::vector<std::vector<node>> outEdges;
```

> Maybe we should consider caching Networkit represenation of edge bundles as well?

Let's see how slow it is first. But caching would definitely make sense.

sphynx/lynxkite-sphynx/networkit_compute_attribute.go
darabos: > Oh, it's been a while since I looked at go code. Really, this is the way you fill a slice with a constant value? _Sigh_

Screenshot 2020-11-17 at 10 49 03

app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
```diff
@@ -0,0 +1,32 @@
...
+      case v: String => json.Json.toJson(v)
+      case v: Int => json.Json.toJson(v)
+      case v: Double => json.Json.toJson(v)
+    }.toSeq)
+    val op = NetworKitComputeAttribute(j.as[json.JsObject])
+    import Scripting._
+    op(op.es, es).result.attr
+  }
+}
+case class NetworKitComputeAttribute(j: json.JsObject)
```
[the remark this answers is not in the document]
darabos: > Can we be just a little bit more typed about this? I understand that we need some arbitrary parameters as well, but can we at least extract the operation name into a proper parameter here?

Done.

sphynx/networkit/additions.h
darabos: > Aha, so all this is ours. But why do we need these two copy steps, isn't one enough? Is it not possible to pass over from go a simple pointer to a pre-allocated slice - hopefully looking like a `*unsigned long long` from he C++ side - and fill things up directly into the slice? You may need to first get `n` over to go in a separate call.

Done. I tried doing it with `SphynxId []` on the C++ side (instead of `SphynxId *`) but that refuses to work. Since I don't even understand why the current solution works, I better stop messing with it.

darabos:
I plan to compile a list of NetworKit algorithms and we can decide where to put them in LynxKite.

xandrew-lynx (commented):
I think we were assuming too much of the go garbage collector cleaning up C++ allocated stuff. It seems it doesn't. This seems like a fairly good summary:

Critical sentence:
_Calling NewClassName for a C++ class ClassName will allocate memory using the C++ memory allocator. This memory will not be automatically freed by Go's garbage collector as the object ownership is not tracked. When you are done with the C++ object you must free it using DeleteClassName._

That said, it's not clear to me how this would explain the weirdness you mentioned. I also don't have another good explanation... :/ Maybe worth poking at that a bit more. E.g. what about writing a stress test to try to draw the error out again? E.g. just doing the current test a million times? Maybe on a hundred threads, a million times each? I'm hoping such a test would be fairly easy to add. If you do this, to test the test, maybe leave in the memory leak first and see if you can a write a test that somehow exposes it.

xandrew-lynx:
I realized I forgot to add the link, sorry:
http://www.swig.org/Doc3.0/Go.html#Go_classes

darabos:
These random generators are lovely.
image

It's very little work to add a new operation. But from time to time there is something new to debug. I still think the overall pace is okay and will allow for pretty wide coverage.

I just found two n-dimensional layout algorithms in NetworKit. Those would go nicely with these random graphs!

app/com/lynxanalytics/biggraph/frontend_operations/GraphComputationOperations.scala
```diff
@@ -85,6 +85,19 @@ class GraphComputationOperations(env: SparkFreeEnvironment) extends ProjectOpera
...
+    def enabled = project.hasEdgeBundle
+    def apply() = {
+      val name = params("name")
+      // A directed graph just gets turned into an undirected graph.
+      // We can skip that and just build an undirected graph.
+      val core = graph_operations.NetworKitComputeAttribute.run(
+        "CoreDecomposition", es, Map("directed" -> false))
+      project.newVertexAttribute(name, core, help)
+    }
+  })
```
xandrew-lynx, line 99: I'm just wondering, shouldn't this create a segmentation instead of an attribute?

sphynx/lynxkite-sphynx/networkit_create_graph.go
```diff
@@ -33,6 +68,23 @@ func init() {
...
+					result = g.Generate()
+				case "Edge switching Markov chain":
+					g := networkit.NewEdgeSwitchingMarkovChainGenerator(degs)
+					defer networkit.DeleteEdgeSwitchingMarkovChainGenerator(g)
+					result = g.Generate()
+				case "Haveli–Hakimi":
+					g := networkit.NewHavelHakimiGenerator(degs)
+					defer networkit.DeleteHavelHakimiGenerator(g)
+					result = g.Generate()
+				}
```
xandrew-lynx, line 87: Shouldn't we also provide these in a form that creates edges only for existing vertices, using a vertex attribute for the requested degree?

xandrew-lynx (commented):
Cool! :) Indeed, the code needed to add a new op looks very reasonable. What is missing, though, and I expect may be a bigger part of the work is testing and documentation. Also, I'm itching to have something merged. :) What about not adding more operations for this PR, instead, finalize the current set with docs and tests and then merge?

app/com/lynxanalytics/biggraph/frontend_operations/GraphComputationOperations.scala
darabos: > I'm just wondering, shouldn't this create a segmentation instead of an attribute?

I don't think it's clear. With an attribute you can immediately put in a filter and get a core, or color a visualization by the core number. What would be easier with a segmentation? Getting the average of some attribute over a core? But that's a bit weird. The vertices in a core are not really related to each other. I don't think there are many cases where you would benefit from that.

But of course there's nothing stopping us from outputting a segmentation in addition to the attribute! Or make it a checkbox. NetworKit outputs the results as a segmentation (Partition) too. I just haven't written the code to turn a NetworKit Partition into a LynxKite segmentation yet. But that's certainly coming.

sphynx/lynxkite-sphynx/networkit_create_graph.go
darabos: > Shouldn't we also provide these in a form that creates edges only for existing vertices, using a vertex attribute for the requested degree?

Ah, that would be pretty cool! Maybe in a follow-up.

darabos:
> What about not adding more operations for this PR, instead, finalize the current set with docs and tests and then merge?

Alright! I've started adding tests. My idea with them is to keep them basic, just to demonstrate that the integration works. The actual algorithms are already tested in NetworKit so just one smoke test for each is enough to confirm that I didn't mess up anything.

Another painful TODO item is to make this work in the CI environment. And for anyone building LynxKite. Because at the moment it just assumes you have checked out and built NetworKit in `~/networkit`. 😰

darabos:
The build is still failing, but now it's because https://maven.twttr.com/ is down. That's almost as good as a passing test. Right? 😅  I think it's a good time to review this!

xandrew-lynx (approved):
Thank you, this looks great!

I agree with your stated testing policy. :)

Also, thanks a lot for the high quality documentation, this really matters a lot!

sphynx/networkit/additions.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"os"
+	"strconv"
+)
+
+// Call this before using NetworKit if you want NETWORKIT_THREADS to apply.
+func SetThreadsFromEnv() {
+	s := os.Getenv("NETWORKIT_THREADS")
+	if t, err := strconv.Atoi(s); err == nil {
```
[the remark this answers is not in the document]
darabos: > Should we really just ignore formatting errors to NETWORKIT_THREADS? I would just panic with a helpful error message.

The most likely error is that it's just not set. 😺  But okay, I can check that separately.

darabos:
✅ ✅ ✅  GREEN TESTS ✅ ✅ ✅ 
Thanks! I'll be back with more!

PR #106, opened 2020-12-02
Layout and community methods from NetworKit

darabos:
Continuing where #102 left off.

I haven't read up on how PivotMDS works, but on small graphs I got silly layouts. Then I tried it on a larger graph:

image

(It's a Mocnik random graph. Left is the default LynxKite layout. Right is PivotMDS. 😍 )

### DIFF
--- a/app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
+++ b/app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
@@ -39,6 +39,27 @@ case class NetworKitComputeAttribute(op: String, options: json.JsObject)
   override def toJson = json.Json.obj("op" -> op, "options" -> options)
 }
 
+object NetworKitComputeVectorAttribute extends OpFromJson {
+  def fromJson(j: json.JsValue) = NetworKitComputeVectorAttribute(
+    (j \ "op").as[String], (j \ "options").as[json.JsObject])
+  def run(name: String, es: EdgeBundle, options: Map[String, Any] = Map())(
+    implicit
+    m: MetaGraphManager): Attribute[Vector[Double]] = {
+    val op = NetworKitComputeVectorAttribute(name, NetworKitCommon.toJson(options))
+    import Scripting._
+    op(op.es, es).result.attr
+  }
+}
+case class NetworKitComputeVectorAttribute(op: String, options: json.JsObject)
+  extends TypedMetaGraphOp[GraphInput, AttributeOutput[Vector[Double]]] {
+  @transient override lazy val inputs = new GraphInput()
+  def outputMeta(instance: MetaGraphOperationInstance) = {
+    implicit val i = instance
+    new AttributeOutput[Vector[Double]](inputs.vs.entity)
+  }
+  override def toJson = json.Json.obj("op" -> op, "options" -> options)
+}
+
 object NetworKitCreateGraph extends OpFromJson {
   def fromJson(j: json.JsValue) = NetworKitCreateGraph(
     (j \ "op").as[String], (j \ "options").as[json.JsObject])
--- a/app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
+++ b/app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
@@ -100,3 +100,36 @@ case class NetworKitCreateGraph(op: String, options: json.JsObject)
   def outputMeta(instance: MetaGraphOperationInstance) = new NetworKitCreateGraph.Output()(instance)
   override def toJson = json.Json.obj("op" -> op, "options" -> options)
 }
+
+object NetworKitCommunityDetection extends OpFromJson {
+  def fromJson(j: json.JsValue) = NetworKitCommunityDetection(
+    (j \ "op").as[String], (j \ "weighted").as[Boolean], (j \ "options").as[json.JsObject])
+  def run(
+    name: String,
+    es: EdgeBundle,
+    options: Map[String, Any] = Map(),
+    weight: Option[Attribute[Double]] = None)(
+    implicit
+    m: MetaGraphManager): Output = {
+    val op = NetworKitCommunityDetection(name, weight.isDefined, NetworKitCommon.toJson(options))
+    import Scripting._
+    weight match {
+      case Some(weight) => op(op.es, es)(op.weight, weight).result
+      case None => op(op.es, es).result
+    }
+  }
+  class Output(implicit
+      instance: MetaGraphOperationInstance,
+      inputs: NetworKitCommon.WeightedGraphInput) extends MagicOutput(instance) {
+    val partitions = vertexSet
+    val belongsTo = edgeBundle(
+      inputs.vs.entity, partitions, properties = EdgeBundleProperties.partialFunction)
+  }
+}
+case class NetworKitCommunityDetection(op: String, weighted: Boolean, options: json.JsObject)
+  extends TypedMetaGraphOp[NetworKitCommon.WeightedGraphInput, NetworKitCommunityDetection.Output] {
+  @transient override lazy val inputs = new NetworKitCommon.WeightedGraphInput(weighted)
+  def outputMeta(instance: MetaGraphOperationInstance) =
+    new NetworKitCommunityDetection.Output()(instance, inputs)
+  override def toJson = json.Json.obj("op" -> op, "weighted" -> weighted, "options" -> options)
+}
--- a/app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
+++ b/app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
@@ -16,48 +16,67 @@ object NetworKitCommon {
       case v: Boolean => json.Json.toJson(v)
     }.toSeq)
   }
+  class WeightedGraphInput(weighted: Boolean) extends MagicInputSignature {
+    val vs = vertexSet
+    val es = edgeBundle(vs, vs)
+    val weight = if (weighted) edgeAttribute[Double](es) else null
+  }
 }
 
 object NetworKitComputeAttribute extends OpFromJson {
   def fromJson(j: json.JsValue) = NetworKitComputeAttribute(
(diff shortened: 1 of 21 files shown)


app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
```diff
@@ -16,48 +16,67 @@ object NetworKitCommon {
...
     implicit
     m: MetaGraphManager): Attribute[Double] = {
-    val op = NetworKitComputeAttribute(name, NetworKitCommon.toJson(options))
+    val op = NetworKitComputeAttribute(name, weight.isDefined, NetworKitCommon.toJson(options))
     import Scripting._
-    op(op.es, es).result.attr
+    weight match {
+      case Some(weight) => op(op.es, es)(op.weight, weight).result.attr
+      case None => op(op.es, es).result.attr
+    }
```
xandrew-lynx, line 41: Nit: `result.attr` after the `}`?

xandrew-lynx (approved):
Thanks!

app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
darabos: > Nit: `result.attr` after the `}`?

Doesn't work just like that:
```
app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala:41: ';' expected but '.' found.
    }.result.attr
     ^
one error found
```
If I have to add an extra variable, this little repetition is probably easier on the eyes.

sphynx/lynxkite-sphynx/networkit_community_detection.go
```diff
@@ -0,0 +1,62 @@
...
+			// The caller can set "directed" to false to create an undirected graph.
+			g := ToNetworKit(vs, es, weight, o.Options["directed"] != false)
+			defer networkit.DeleteGraph(g)
+			var p networkit.Partition
+			switch ea.GetStringParam("op") {
+			case "PLM":
+				c := networkit.NewPLM(g, false, o.Double("gamma"))
+				defer networkit.DeletePLM(c)
+				c.Run()
+				p = c.GetPartition()
```
darabos, line 38: I've got to tell someone how I wasted half a day on a single-character mistake.
```suggestion
				p := c.GetPartition()
```
With `:=` instead of `=` this makes a local variable instead of modifying `p` outside of the `switch`. So `p.NumberOfSets()` below fails. It actually carries the `nil` to the C++ code and segfaults there. The SWIG wrapper catches that somehow (?) and we `recover` due to the code above. The code is supposed to return an error then. Something's wrong and instead of causing an error on the frontend, something causes this `execute` function to be called again. (Didn't look into why this happens.)

So PLM just gets run again and again until we somehow end up dying of stack overflow. (Probably related to the unexpected retries.)

The PLM C++ code outputs some debug info. So I just see that for thousands of lines and then stack overflow. PLM has a `recurse` parameter that defaults to `true`. Must be an infinite recursion! I read the PLM code and when I can't figure out how it could do infinite recursion, I recompile it with tons of extra logging.

After a while I figured out it wasn't infinite recursion in PLM but a segfault in `p.numberOfSubsets()`. `PLM::getPartition()` returns a `Partition` by value instead of pointer. SWIG [wraps it by copying the returned value](http://www.swig.org/Doc4.0/SWIG.html#SWIG_nn23) and I became convinced that this copy holds some pointer to memory that the original Partition frees. I looked at the wrapper code with suspicion:

```c++
  *(Partition **)&_swig_go_result = new Partition(result);                                          
```

What is all that? I wrote my own wrapper with a simpler interface. Once I managed to get that to compile and it didn't help at all, I soon realized that I wrote `:=` instead of `=`.

😭 

But check out these cool communities!

image

darabos:
Sorry, I'll add a few more community detection methods and then add docs & tests and that's it for this PR.

app/com/lynxanalytics/biggraph/graph_operations/NetworKit.scala
xandrew-lynx: Yeah, don't introduce an extra variable for sure. It doesn't really matter, totally fine to leave as is, but maybe it would work if you did `(weight match { ... }).result.attr`?

sphynx/lynxkite-sphynx/networkit_community_detection.go
xandrew-lynx: Wow, this is a very sad story. :( But wow, this `:=` vs `=` seems like a terrible language design. I thought this is the kind of problems you _don't_ get in go due to the the radically reduced set of language features. Oh well. :/

darabos:
I've added two more methods in one box.

image

There's one more method in NetworKit (`ParallelAgglomerativeClusterer`) but it unfortunately doesn't work well. On the random graphs and example graph it fails with [`G has self-loops and cannot be processed`](https://github.com/networkit/networkit/blob/7.1/networkit/cpp/matching/PathGrowingMatcher.cpp#L15) even if I run `g.removeSelfLoops()`. It's not exposed in Python and there's a [TODO](https://github.com/networkit/networkit/blob/7.1/networkit/cpp/community/test/CommunityGTest.cpp#L467) to overhaul it. So let's skip it for now.

Docs and tests are next!

sphynx/lynxkite-sphynx/networkit_compute_attribute.go
```diff
@@ -34,7 +34,6 @@ func init() {
 				Defined: make([]bool, len(vs.MappingToUnordered)),
 			}
 			var result networkit.DoubleVector
-			defer networkit.DeleteDoubleVector(result)
```
xandrew-lynx: So wait, then we don't need to delete the "right thing", either?

web/app/help/operations/place-vertices-with-edge-lengths.asciidoc
```diff
@@ -0,0 +1,48 @@
+### Place vertices with edge lengths
+
+These methods create a graph layout as a new ``Vector[number]`` vertex attribute
+where the edges have given lengths.
```
xandrew-lynx, line 4: Shouldn't this be worded a bit weaker? ... in a way which makes the physical distance between connected vertices be as close as possible to given edge length? Or something better. :) I mean, this sounds like we find a perfect solution.

xandrew-lynx (approved):
Thanks, great!

sphynx/lynxkite-sphynx/networkit_community_detection.go
```diff
@@ -0,0 +1,62 @@
...
+			vs.MappingToUnordered = make([]int64, p.NumberOfSubsets())
+			for i := range vs.MappingToUnordered {
+				vs.MappingToUnordered[i] = int64(i)
+			}
+			es = &EdgeBundle{}
+			es.EdgeMapping = make([]int64, p.NumberOfElements())
+			es.Src = make([]SphynxId, p.NumberOfElements())
+			es.Dst = make([]SphynxId, p.NumberOfElements())
+			v := p.GetVector()
+			defer networkit.DeleteUint64Vector(v)
```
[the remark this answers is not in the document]
darabos: > I'm a bit confused by this. Why do we need to delete this separately? Isn't this part of Partition which is deleted via the above deferred? Networkit seems to just return a reference to a non-pointer member:
> https://github.com/networkit/networkit/blob/b354604aad07ce9ff3ef195e6dadcfc313818308/networkit/cpp/structures/Partition.cpp#L132
> (The comment on this implementation is not exactly reassuring, though. :) )

I think that `return` would cause a copy.

> Or wait. It is not returning a reference, but directly a vector. So is calling getVector causes a copy? Sorry, my C++ memories are fading. But that seems terrible. Wouldn't the normal way be returning a const reference?

It's a SWIG thing. It uses the copy constructor when a function returns something by value: http://www.swig.org/Doc4.0/SWIG.html#SWIG_nn23

Since the `vector` value is not used again after that, I think the move constructor will probably be invoked. But still the end result is that we own the pointer.

I'm super unsure about all this. I've asked on Stack Overflow but didn't get a response. https://stackoverflow.com/questions/65112791/sanitizing-memory-management-in-go-swig-c-code

Currently I mostly try to delete stuff, and if I get an error about double deletes I remove it. 😅  Like a total pro.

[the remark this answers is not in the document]
xandrew-lynx: Right, I'm totally convinced now. 👍

But seriously, I think it's probably indeed the right thing based on the SWIG doc you pointed to.

That said, it's also an option to opt for something less magical, like what you did for edges. I.e., create something on the C++ side that takes a vector pointer and copies the stuff into it. Then ownership is clear.

sphynx/lynxkite-sphynx/networkit_compute_attribute.go
darabos: > So wait, then we don't need to delete the "right thing", either?

This tries to delete `nil`. If I move this line after the `switch`, it crashes with `free(): invalid pointer`. I took that to mean we don't need to free it. But I have no idea. 😓

web/app/help/operations/place-vertices-with-edge-lengths.asciidoc
darabos: > Shouldn't this be worded a bit weaker?

Done, thanks.

sphynx/lynxkite-sphynx/networkit_community_detection.go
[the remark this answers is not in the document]
darabos: > That said, it's also an option to opt for something less magical, like what you did for edges. I.e., create something on the C++ side that takes a vector pointer and copies the stuff into it. Then ownership is clear.

I don't like that that uses the `unsafe` Go API and I'm afraid it would need a bit of extra code for each class where we want to do this. Letting SWIG do it's thing seems less magical to me than trying to implement the magic ourselves. 😄 

I feel like it's too early to try to write good SWIG code, since we are such noobs. But one day! We will understand everything and solve these things nicely. Until then I'm happy if things just work.

PR #111, opened 2020-12-10
NetworKit integration: scalars, fixes, and an edge score

darabos:
I'll add more to this.


sphynx/lynxkite-sphynx/networkit_compute_double_edge_attribute.go
```diff
@@ -0,0 +1,59 @@
...
+				c.Run()
+				result = c.Scores()
+			}
+			attr := &DoubleAttribute{
+				Values:  make([]float64, len(es.Src)),
+				Defined: make([]bool, len(es.Src)),
+			}
+			for i := range attr.Values {
+				// The NetworKit edge IDs don't correspond to the Sphynx edge IDs.
+				// TODO: Can we do better than mapping the results back by src/dst?
```
darabos, line 51: This sucks.

NetworKit edges doesn't normally have IDs because they don't really need to be identified. Except in these edge scoring algorithms. So they require you to run `indexEdges()` first. `indexEdges()` just hands out edge IDs sequentially from 0. Ideally we would have the same edge IDs in NetworKit as in Sphynx, so the returned scores could be applied directly. But some other mapping could work too. I didn't really find anything that would work, and now have to loop up the edges by src/dst.

One idea is to put the Sphynx ID into the weight when I build the graph. Then call `indexEdges()` and we can read out the mapping. But I can't find a way to then replace these fake weights with the real weights. And anyway it would be pretty ugly.

Let me know if you have better ideas.

xandrew-lynx: Why is this current solution so bad? Are you worried that `g.EdgeId` is slow?

Somewhat related: what's the deal with parallel edges? As far as it seems to me now, NetworKit does not care for parallel edges. So there can be only one weight for an (a, b) pair and there will only be one result as well. Do we handle this correctly when we convert the weights for NK? Because of this, we cannot hope for a 1 to 1 correspondence between our ids and networKit ids, can we?

darabos: > Why is this current solution so bad? Are you worried that `g.EdgeId` is slow?

Yes. It looks up `u` in `outEdges` (O(1)) and then scans through the outgoing edge list to find `v` (O(`outEdges[u].size`)). Knowing its index it can look up the ID in `outEdgeIds` (O(1)).

```c++
    std::vector<std::vector<node>> outEdges;
    std::vector<std::vector<edgeid>> outEdgeIds;
    std::vector<std::vector<edgeweight>> outEdgeWeights;
```

> Somewhat related: what's the deal with parallel edges?

Yeah, this will mix up the parallel edges. We will output the score of the "first" edge for all parallel edges.

NetworKit can represent parallel edges. You just have the same neighbor twice in `outEdges[u]`. Some algorithms may not handle them well, as they may look up edge weights by src/dst which will just return the first weight.

xandrew-lynx: I don't seem to find it, but in case I'm just missing it: is there a (hopefully O(1)) way to get src and dst for a given NK edge id? If yes, then worse case you can just use this to build a (src, dst) -> NK Id map on your end and use this to avoid the O(outdegree) cost. But even if possible, I'd defer that to a separate issue for now. I don't have any better idea than this unfortunately.

sphynx/lynxkite-sphynx/networkit_test.go
```diff
@@ -37,7 +37,13 @@ func TestNewVertexAttribute(t *testing.T) {
 	b := networkit.NewBetweenness(ExampleGraph())
 	defer networkit.DeleteBetweenness(b)
 	b.Run()
-	s := ToDoubleSlice(b.Scores())
+	v := b.Scores()
+	// TODO: Understand why this delete fails. Do we leak this vector?
+	// defer networkit.DeleteDoubleVector(v)
```
xandrew-lynx, line 42: Please create an issue for this, so that we don't forget.

web/app/help/operations/compute-score-edges-with-the-forest-fire-model.asciidoc
```diff
@@ -0,0 +1,32 @@
+### Compute degree
```
xandrew-lynx, line 1: This doesn't seem right. :)

xandrew-lynx (approved):
Lots of great stuff, thank you!

sphynx/lynxkite-sphynx/networkit_compute_double_edge_attribute.go
darabos: > I don't seem to find it, but in case I'm just missing it: is there a (hopefully O(1)) way to get src and dst for a given NK edge id? If yes, then worse case you can just use this to build a (src, dst) -> NK Id map on your end and use this to avoid the O(outdegree) cost. But even if possible, I'd defer that to a separate issue for now. I don't have any better idea than this unfortunately.

Not exactly from ID to src/dst, but we can look up both together! This should work! Thanks. I think I'll just do it now. I won't remember anything by next year.

```c++
    std::pair<node, edgeid> getIthNeighborWithId(node u, index i) const {
```

xandrew-lynx: This API looks like a real pleasure to work with. :) But OK, if you do it, that's great!

darabos:
The frontend test failure looks like a fluke. (`Failed: No element found using locator: By(css selector, *[id="new-directory"])
`) It has passed before. I'll merge this!

web/app/help/operations/find-optimal-spanning-tree.asciidoc
```diff
@@ -0,0 +1,28 @@
+### Find optimal spanning tree
```
darabos, line 1: But please take a look at this. I've added this whole box after your approval. Thanks!

web/app/help/operations/find-optimal-spanning-tree.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+(or highest) possible total edge weight.
+
+Uses the
+https://networkit.github.io/dev-docs/cpp_api/classNetworKit_1_1RandomMaximumSpanningForest.html[NetworKit]
+implementation.
+
+====
+[p-name]#Save as#::
+The new edge attribute will be created under this name. Edges that make up the
+tree are marked by 1 and the edges that are not part of the tree are marked by 0.
```
xandrew-lynx, line 15: I wonder if 1 and undefined would be better? It would be less space and I think we do stuff like that in other operations. Maybe less convenient, though? Not sure.

darabos: > I wonder if 1 and undefined would be better? It would be less space and I think we do stuff like that in other operations. Maybe less convenient, though? Not sure.

I wasn't sure either. I don't see much precedent. "Add reversed edges" uses 0 and 1.

For visualization undefined is a bit worse because it will be the default blue and the color assigned to 1 will not contrast with it in many color scales.

For filtering they are both fine.

For derived attributes and Python use... I'm not sure.

But you're right about the storage on the Spark side! I've sent a PR to switch to undefined/1: #118. I'll leave it for you to decide whether to merge it.

xandrew-lynx: (To keep the discussion in one place let me briefly answer here.)

I'm torn because of the visualization thing and also if you look at things with operations research eyes then 0 is more logical, but still, I'll go ahead and merge that PR. Mostly for consistency. Otherwise I wouldn't have a good way to explain why the find Steiner tree op works differently. And also we use this undefined convention in inputs as well, e.g. shortest paths and well, Steiner tree again.

PR #115, opened 2020-12-14
Avoid infinite error loop

### LINKED ISSUE #103: Clear error reporting when LynxKite is down
I got this screenshot from a user:
image

I'm pretty sure LynxKite is down. This causes some requests to fail, which is not handled correctly by some piece of JavaScript, so there's a JavaScript error, which we try to send to the backend for logging. But that request fails too, and this time we handle it by popping up these alerts.

- Let's not report `ajax/jsError` failures. It's never useful.
- Make sure we display a clear error when LynxKite is not accessible from the frontend.

darabos:
Fixes #103. I just hit the same problem from a misconfiguration. The infinite errors were really unhelpful and even killed Chrome Dev Tools. This change allowed me to see the real error.

### DIFF
--- a/web/app/help/operations/compute-centrality.asciidoc
+++ b/web/app/help/operations/compute-centrality.asciidoc
@@ -10,58 +10,58 @@ The new attribute will be created under this name.
 
 [p-algorithm]#The centrality algorithm to use#::
 - **Average distance**
-  (or https://en.wikipedia.org/wiki/Centrality#Closeness_centrality[closeness centrality])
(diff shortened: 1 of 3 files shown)


darabos:
Sorry, I threw in a small fix for the documentation. The `+` has to be at the start of the line to work as a [list continuation](https://asciidoctor.org/docs/asciidoc-writers-guide/#list-continuation).

xandrew-lynx (approved):
Thanks for the fix!

PR #118, opened 2020-12-16
Undefined instead of 0 in spanning tree

darabos:
For https://github.com/lynxkite/lynxkite/pull/111#discussion_r544229741.


xandrew-lynx (approved):
Is the test failure a red herring?

darabos:
> Is the test failure a red herring?

No I think it's real. I'll fix it.

PR #123, opened 2021-01-04
Segmentation metrics from NetworKit

darabos:
There are 7 more per-segment metrics like this one. One of them takes two segmentations as input. I think I'll skip that one and just add the 6 that have the same interface.

There are also 5 segmentation metrics that are just a single scalar for a whole segmentation. An example is _modularity_. (I originally missed these because they don't derive from the `Algorithm` class.) I'll add these too.

I would be fine putting these all into the new "Segmentation attributes" box category. Or do you have a better idea for organization?

Also not sure about separate boxes vs one box with a dropdown. But I like separate boxes. It leaves more room for documentation, you more easily find them in the box search, saves the user from picking from a dropdown. So I'll go that way if you don't stop me.


darabos:
It was all pretty smooth so far, but conductance is always 1 for some reason. I'm debugging that now. TODO: tests and the scalar metrics.

sphynx/lynxkite-sphynx/networkit_compute_segment_attribute.go
```diff
@@ -0,0 +1,120 @@
...
+				}
+			}
+
+			// The caller has to call networkit.DeletePartition.
+			partition := func() networkit.Partition {
+				p := networkit.NewPartition(uint64(len(vs.MappingToUnordered)))
+				p.SetUpperBound(uint64(len(seg.MappingToUnordered)))
+				log.Printf("counts: %v %v %v", len(vs.MappingToUnordered), len(seg.MappingToUnordered))
+				for i := range belongsTo.EdgeMapping {
+					log.Printf("%v %v %v", i, belongsTo.Dst[i], belongsTo.Src[i])
```
xandrew-lynx, line 54: Remove unnecessary logs, also above and below.

web/app/help/operations/compute-segment-conductance.asciidoc
```diff
@@ -0,0 +1,25 @@
+### Compute segment conductance
+
+Computes the conductance of each segment in a non-overlapping segmentation.
```
xandrew-lynx, line 3: Shouldn't we enforce the non-overlappingness of the segmentation somewhere in the code? Probably the frontend operation should check the edge bundle properties of the belongs to bundle.

web/app/help/operations/compute-segment-expansion.asciidoc
```diff
@@ -0,0 +1,25 @@
+### Compute segment expansion
+
+Computes the expansion of each segment in a non-overlapping segmentation.
```
xandrew-lynx, line 3: Interesting that all these ops need non-overlapping segments. The definitions don't seem to need it. I guess that's for some implementation shortcuts?

xandrew-lynx (commented):
Cool, thanks, interesting operations, looks good so far!

app/com/lynxanalytics/biggraph/frontend_operations/Operations.scala
```diff
@@ -48,7 +49,7 @@ object Categories {
   val BuildSegmentationOperations =
     Category("Build segmentation", "green", icon = "th-large", index = idx)
   val UseSegmentationOperations =
-    Category("Use segmentation", "orange", icon = "th-large", index = idx)
+    Category("Use segmentation", "blue", icon = "th-large", index = idx)
```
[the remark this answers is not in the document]
darabos: > Not that I mind, but why the color change?

The vertex and edge attribute categories are orange. So I wanted the segmentation attribute category to be orange too. I stole the orange from here. 😼

web/app/help/operations/compute-segment-expansion.asciidoc
darabos: > Interesting that all these ops need non-overlapping segments. The definitions don't seem to need it. I guess that's for some implementation shortcuts?

They take a `Partition` as the input. The `Partition` is non-overlapping and the `Cover` is overlapping in NetworKit. (I think? Correct me if you think that's not how they work.)

xandrew-lynx: Oh yeah, I think you capture the networkit API correctly, my musing were more about why _they_ decided to work with a partition and not a cover and I meant it must be an implementation shortcut on _their_ end. No need to spend time to investigate anything, it is what it is, just wanted to point out that it's a bit curious.

darabos:
> It was all pretty smooth so far, but conductance is always 1 for some reason. I'm debugging that now.

It was a NetworKit bug! https://github.com/networkit/networkit/pull/647

darabos:
image

These are the boxes in the end. Maybe hub dominance should be merged into one box? They are two separate pieces in NetworKit, but the global metric is just the average of the per-segment version.

It's ready for a last review! I'm not adding anything more and I've addressed your comments so far. Thanks!

test/com/lynxanalytics/biggraph/frontend_operations/NetworKitTest.scala
```diff
@@ -32,7 +32,7 @@ class NetworKitTest extends OperationsTestBase {
         "Betweenness (estimate)" -> Map(0 -> 0.0, 1 -> 0.0, 2 -> 0.0, 3 -> 0.0),
         "Eigenvector" -> Map(0 -> 0.7, 1 -> 0.7, 2 -> 0.0, 3 -> 0.0),
         "Harmonic Closeness" -> Map(0 -> 0.33, 1 -> 0.33, 2 -> 0.66, 3 -> 0.0),
-        "Katz" -> Map(0 -> 0.5, 1 -> 0.5, 2 -> 0.49, 3 -> 0.49),
+        "Katz" -> Map(0 -> 0.7, 1 -> 0.7, 2 -> 0.06, 3 -> 0.06),
```
xandrew-lynx, line 36: So I guess this is the line that the commit title refers to? The rest is just minor refactor not worthy of a mention... ;)

web/app/help/operations/compute-modularity-of-segmentation.asciidoc
```diff
@@ -0,0 +1,25 @@
+### Compute modularity of segmentation
```
xandrew-lynx, line 1: We already have a backend operation that computes this and we expose its result when we use find modular clustering. (https://github.com/lynxkite/lynxkite/blob/master/app/com/lynxanalytics/biggraph/frontend_operations/BuildSegmentationOperations.scala#L254) This is a bit suboptimal. But I guess we can leave with it if at least the results are the same. But it would be very bad if computing the modularity with this operation on the result of a find modular clustering op yielded a different number. Can you check this? Also, we can consider using this operation instead of the old one in that FE operation. Somewhat related, doesn't networkit has a better find modular clustering implementation? Maybe we should replace that, too?

test/com/lynxanalytics/biggraph/frontend_operations/NetworKitTest.scala
xandrew-lynx: It's fine of course and I do like this "minor refactor" a lot.

xandrew-lynx (approved):
Cool! Indeed, I agree that we should unify or even just remove the segmentation scalar version of the hub dominance operation. (I mean, if it's just an average maybe we can let the user compute it if needed.)

test/com/lynxanalytics/biggraph/frontend_operations/NetworKitTest.scala
darabos: > So I guess this is the line that the commit title refers to? The rest is just minor refactor not worthy of a mention... ;)

Huh, good catch! I thought I committed this separately? Fixed in a rebase now.

web/app/help/operations/compute-modularity-of-segmentation.asciidoc
darabos: > We already have a backend operation that computes this and we expose its result when we use find modular clustering.

Oh I didn't realize that! Why didn't we expose it as a box...?

> This is a bit suboptimal.

We have Sphynx implementations of a bunch of Spark operations. What is not to like?

> But I guess we can leave with it if at least the results are the same. But it would be very bad if computing the modularity with this operation on the result of a find modular clustering op yielded a different number. Can you check this?

image

Is that close enough? 😸 

Looking at the [NetworKit](https://github.com/networkit/networkit/blob/8.0/networkit/cpp/community/Modularity.cpp) and [LynxKite](https://github.com/lynxkite/lynxkite/blob/4.1.0/app/com/lynxanalytics/biggraph/graph_operations/Modularity.scala) code, I thought the only difference is that we consider the in and out degrees separately, while they just look at the total degree and assume it's 50/50 in/out. But even when I change our code to do the same, the results don't match. I guess I'll debug this a bit.

> Somewhat related, doesn't networkit has a better find modular clustering implementation? Maybe we should replace that, too?

Your implementation is based on the Louvain method, right? That is already integrated as the "Find communities with the Louvain method" box. I don't know if it's compatible enough to be a replacement.

darabos:
> Cool! Indeed, I agree that we should unify or even just remove the segmentation scalar version of the hub dominance operation. (I mean, if it's just an average maybe we can let the user compute it if needed.)

👍  Done.

web/app/help/operations/compute-modularity-of-segmentation.asciidoc
xandrew-lynx: > We have Sphynx implementations of a bunch of Spark operations. What is not to like?
Yes, but normally they are just different implementations of the same abstract operation that supposed to do the same thing.

> Your implementation is based on the Louvain method, right?
Well, I would think not, at least not exactly. It has some not too pretty stuff to do things in a distributed way and there is no hierarchical aspect in my implementation.

michitux: > Looking at the [NetworKit](https://github.com/networkit/networkit/blob/8.0/networkit/cpp/community/Modularity.cpp) and [LynxKite](https://github.com/lynxkite/lynxkite/blob/4.1.0/app/com/lynxanalytics/biggraph/graph_operations/Modularity.scala) code, I thought the only difference is that we consider the in and out degrees separately, while they just look at the total degree and assume it's 50/50 in/out. But even when I change our code to do the same, the results don't match. I guess I'll debug this a bit.

NetworKit just implements Modularity for undirected graphs. An equivalent directed graph contains for every undirected edge `{u, v}` two edges `(u, v)` and `(v, u)`. Thus, the undirected degree is the same as the in-degree and the out-degree and the number of edges in the directed graph is twice the number of edges in the undirected graph. I hope that helps with debugging. I would be surprised if the NetworKit version was wrong, but I of course cannot fully exclude that.

Regarding distributed modularity-based clustering, are you aware of our work at https://github.com/kit-algo/distributed_clustering_thrill? It is implemented in thrill, but from the few source files I have seen so far, the algorithmic ideas should be directly translatable into your framework. If you want high quality, I suggest having a look at the Leiden algorithm, an extension of the Louvain algorithm (see, e.g., https://github.com/vtraag/leidenalg). With some luck (I have a student working on it), we might get an implementation of Leiden in NetworKit till the end of February.

web/app/help/operations/compute-segment-expansion.asciidoc
michitux: As I originally implemented these measures that derive from `LocalPartitionEvaluation`, I will try answering this question. First, I did not need them for overlapping communities/segments as you call them, just for non-overlapping ones.  Second, if I recall correctly, it was also not clear to me how they should be generalized to overlapping communities/segments but looking at them again now I do not see why there should be any issue at least with measures like conductance and expansion that are basically just defined for single communities/segments. As a cover stores a separate data structure for every node that contains the ids of all communities it is part of, this is quite some overhead compared to the single integer that is stored for a partition which is why converting everything to a cover might not always be a good idea. Nevertheless, I agree that it would be good to implement them for overlapping instead of disjoint communities and support disjoint communities, e.g., via some adapter. I created networkit/networkit#650 to document these ideas in NetworKit.

xandrew-lynx: Thanks for the explanation Michael, makes perfect sense! We do not have any pressing need right now to apply these on overlapping communities, but of course would be great to have what you propose in https://github.com/networkit/networkit/issues/650, thanks for creating it. So no hurry from our end, but we will certainly extend our boxes as well to overlaps if that is done.

web/app/help/operations/compute-modularity-of-segmentation.asciidoc
darabos: > > Looking at the [NetworKit](https://github.com/networkit/networkit/blob/8.0/networkit/cpp/community/Modularity.cpp) and [LynxKite](https://github.com/lynxkite/lynxkite/blob/4.1.0/app/com/lynxanalytics/biggraph/graph_operations/Modularity.scala) code, I thought the only difference is that we consider the in and out degrees separately, while they just look at the total degree and assume it's 50/50 in/out. But even when I change our code to do the same, the results don't match. I guess I'll debug this a bit.
> 
> NetworKit just implements Modularity for undirected graphs.

Thanks! You're right, when I give it an undirected graph I get the same number as in LynxKite.

@xandrew-lynx: So now we have two ways to compute the same scalar. I prefer the new API because it made it easy to add all the metrics the same way. And I don't mind leaving the old LynxKite Modularity op as it is. It's only used in "Find modular clustering", which is in Spark anyway — I don't think we would gain much by moving the whole graph to Sphynx just for this scalar.

> Regarding distributed modularity-based clustering, are you aware of our work at https://github.com/kit-algo/distributed_clustering_thrill? It is implemented in thrill, but from the few source files I have seen so far, the algorithmic ideas should be directly translatable into your framework. If you want high quality, I suggest having a look at the Leiden algorithm, an extension of the Louvain algorithm (see, e.g., https://github.com/vtraag/leidenalg). With some luck (I have a student working on it), we might get an implementation of Leiden in NetworKit till the end of February.

Thanks for the links! These are all fantastic!

PR #128, opened 2021-01-18
Anonymous data collection

darabos:
I tried logging to Firestore first. But I couldn't find a way to do streaming analytics on the data that way. It can export to BigQuery but only with an explicit action. With Firebase Analytics on the other hand I got the data streaming into BigQuery where we can analyze it to our heart's content.

For now it only logs 3 events:
- Every backend request, including the endpoint, the duration, and the error if it failed.
- When a box is added, including the box type. (Custom boxes are not logged.)
- When a vertex or edge count is loaded, the number rounded to two significant digits.

image

### DIFF
--- a/web/app/scripts/workspace/workspace-wrapper.js
+++ b/web/app/scripts/workspace/workspace-wrapper.js
@@ -237,7 +237,11 @@ angular.module('biggraph')
         if (!opts.willSaveLater) {
           this.saveWorkspace();
         }
-        util.logUsage('box added', { operationId });
+        const meta = this._boxCatalogMap[operationId];
+        if (meta && meta.categoryId !== 'Custom boxes') {
+          // Collect usage statistics for standard boxes.
+          util.logUsage('box added', { operationId });
+        }
         return box;
       },
 
--- a/web/app/scripts/workspace/workspace-wrapper.js
+++ b/web/app/scripts/workspace/workspace-wrapper.js
@@ -238,10 +238,9 @@ angular.module('biggraph')
           this.saveWorkspace();
         }
         const meta = this._boxCatalogMap[operationId];
-        if (meta && meta.categoryId !== 'Custom boxes') {
-          // Collect usage statistics for standard boxes.
-          util.logUsage('box added', { operationId });
-        }
+        // Collect usage statistics for placed boxes.
+        const loggedId = (meta && meta.categoryId !== 'Custom boxes') ? operationId : 'custom';
+        util.logUsage('box added', { operationId: loggedId });
         return box;
       },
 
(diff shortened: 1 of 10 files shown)


web/app/scripts/workspace/workspace-wrapper.js
```diff
@@ -237,6 +237,10 @@ angular.module('biggraph')
         if (!opts.willSaveLater) {
           this.saveWorkspace();
         }
+        const meta = this._boxCatalogMap[operationId];
+        // Collect usage statistics for placed boxes.
+        const loggedId = (meta && meta.categoryId !== 'Custom boxes') ? operationId : 'custom';
+        util.logUsage('box added', { operationId: loggedId });
```
xandrew-lynx, line 243: Maybe just don't send anything for custom boxes? Or change the document explaining the data collection. When I read there that "Custom boxes are excluded" I would have expected a behavior that they are not logged at all.

web/gulpfile.js
```diff
@@ -188,7 +188,7 @@ gulp.task('serve', gulp.series('quick', function serve() {
       });
   });
   gulp.watch('app/styles/*.css', gulp.series('css'));
-  gulp.watch('app/scripts/**/*.js', gulp.series('eslint', 'js'));
+  gulp.watch('app/scripts/**/*.js', gulp.parallel('eslint', 'js'));
```
xandrew-lynx, line 191: Can you explain a bit what's going on in this file and `basic-annotate.js`? I have no idea why you did these changes. Or is it just some API change?

xandrew-lynx (approved):
Very cool, thank you!

web/app/scripts/workspace/workspace-wrapper.js
darabos: > Maybe just don't send anything for custom boxes? Or change the document explaining the data collection. When I read there that "Custom boxes are excluded" I would have expected a behavior that they are not logged at all.

When I wrote that we didn't log anything for custom boxes. But then I noticed in the design doc we planned to log "custom box" for custom boxes. It makes sense — this way we see how often custom boxes are used.

If you agree, I'll update the policy page and keep the code like this.

web/gulpfile.js
darabos: > Can you explain a bit what's going on in this file and `basic-annotate.js`? I have no idea why you did these changes. Or is it just some API change?

`gulp serve` was crashing on every change. I guess it was like that since I replaced `ng-annotate` with `basic-annotate`. One mistake was that instead of a new `through.obj` it kept using the same one. I think adding `async` was required due to a Gulp API change.

PR #129, opened 2021-01-19
More feedback, less failing for drag & drop uploads

### LINKED ISSUE #96: "Don't move the box!"
Don't allow users do anything while an upload is in progress.

darabos:
Fixes #96. Instead of trying to lock everything up for the duration of the upload, it's more of a proper fix. The issue was that the box object does not survive a `setAndGetWorkspace` request. A new box object with the same ID takes its place. But the fix is simple, we just need to look up the right box by ID.

While there I also added a progress indicator for the upload.

### DIFF
--- a/web/app/scripts/workspace/workspace-drawing-board.js
+++ b/web/app/scripts/workspace/workspace-drawing-board.js
@@ -684,25 +684,35 @@ angular.module('biggraph').directive(
             op = 'Import ORC';
           }
           const box = scope.workspace.addBox(op, event, { willSaveLater: true });
+          box.parameters.filename = 'uploading...';
           if (op === 'Import CSV') {
             box.parameters.infer = 'yes';
           }
-          uploadFile(file).then(function(filename) {
-            box.parameters.filename = filename;
+          uploadFile(file, function(percent) {
+            const b = scope.workspace.boxMap[box.id];
+            scope.$apply(() => {
+              b.summary = `Uploading... (${percent}%)`;
+            });
+          }).then(function(filename) {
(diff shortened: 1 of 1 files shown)


web/app/scripts/workspace/workspace-drawing-board.js
```diff
@@ -684,25 +684,35 @@ angular.module('biggraph').directive(
...
           }
           const box = scope.workspace.addBox(op, event, { willSaveLater: true });
+          box.parameters.filename = 'uploading...';
           if (op === 'Import CSV') {
             box.parameters.infer = 'yes';
           }
-          uploadFile(file).then(function(filename) {
-            box.parameters.filename = filename;
+          uploadFile(file, function(percent) {
+            const b = scope.workspace.boxMap[box.id];
```
xandrew-lynx, line 692: Now the whole `box` object survives thanks to the closure. (It's just not the right box object after a reload, if I understand this correctly.) But in this new implementation all we need is the id as we will do a lookup for the fresh box object. So what about saving the id in a separate variable and only capturing that in the closure?

xandrew-lynx (approved):
Brilliant!

web/app/scripts/workspace/workspace-drawing-board.js
darabos: > Now the whole `box` object survives thanks to the closure. (It's just not the right box object after a reload, if I understand this correctly.) But in this new implementation all we need is the id as we will do a lookup for the fresh box object. So what about saving the id in a separate variable and only capturing that in the closure?

I'm not terribly worried about the memory use of that small object during an upload. Should I be?

PR #133, opened 2021-01-20
Clear error for empty tables

### LINKED ISSUE #119: No "vertices" table when there are no vertex attributes
The `vertices` table does not exist, but the `edges` table does.

sql-on-networkit-graph

My first guess is that the "Create Erdős–Rényi graph" operation runs outside the Spark domain (I'm not sure it's the correct terminology), so the `vertices` table is not created automatically for the Spark SQL version.

After running an edge attribute computation, the `edges` table is created properly and I can run SQL on it.

edges-of-networkit-graph

If I add a "Compute degree" operation, the `vertices` table appears.

darabos:
Fixes #119.

| Before | After |
|---|---|
|image|image|

Now it fails before execution.

### DIFF
--- a/app/com/lynxanalytics/biggraph/controllers/ProtoTable.scala
+++ b/app/com/lynxanalytics/biggraph/controllers/ProtoTable.scala
@@ -18,8 +18,12 @@ import org.apache.spark.sql.catalyst.plans.logical.UnaryNode
 import org.apache.spark.sql.catalyst.plans.logical.Union
 import org.apache.spark.sql.catalyst.plans.logical.Filter
 import org.apache.spark.sql.catalyst.plans.logical.Join
-
 import org.apache.spark.sql.types
+
+class ProtoLocalRelation(val proto: ProtoTable, fields: Seq[types.StructField])
+  extends LocalRelation(fields.map {
+    f => AttributeReference(f.name, f.dataType, f.nullable, f.metadata)()
+  })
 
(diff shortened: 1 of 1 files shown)


app/com/lynxanalytics/biggraph/controllers/ProtoTable.scala
```diff
@@ -38,11 +42,7 @@ trait ProtoTable {
     // unique to the ProtoTable. This breaks a couple of things down the line, so we have to strip
     // it out when this relation is used anywhere else, which might not be trivial.
     val fields = schema.fields.map(f => f.withComment(s"$this"))
-    if (fields.nonEmpty) {
-      LocalRelation(fields.head, fields.tail: _*)
-    } else {
-      LocalRelation()
-    }
+    new ProtoLocalRelation(this, fields)
```
darabos, line 45: This is probably a better solution than the comment. But I didn't dare mess with this scary code. Should I?

xandrew-lynx (approved):

app/com/lynxanalytics/biggraph/controllers/ProtoTable.scala
xandrew-lynx: Just for the record. We concluded it would most probably work to stop using comments altogether and take the proto tables directly from the `ProtoLocalRelation`s. But we are not 100% sure that it's impossible that in some cases the ProtoLocalRelation get converted into some other kind of `LogicalPlan` in which case we lose this information.

PR #134, opened 2021-01-20
Still log to rolling log files in Docker

### LINKED ISSUE #126: Save log files even in Docker
In Docker mode LynxKite only logs to stderr which goes to the Docker logs. LynxKite has a `/logs` page for accessing the log files, but in Docker mode there's nothing there. This makes the log files harder to access, not easier. How about we log to both stderr and the file?

darabos:
Resolves #126. The logs are accessible from `/logs` in a browser.


xandrew-lynx:
LGTM

PR #138, opened 2021-01-21
Properly forward error code in chroot

### LINKED ISSUE #92: Python error handling
In case of errors we should
- print all output, so we can add print() calls for debugging
- handle it right for the "Create graph in Python" box too.

darabos:
Fixes #92. Some context: https://stackoverflow.com/questions/808541

I've tried this locally in a chrooted setup and it got me the same behavior as outside of chroot.

### DIFF
--- a/sphynx/python/util.py
+++ b/sphynx/python/util.py
@@ -177,8 +177,9 @@ class Op:
       os.chdir('/')
       self.datadir = '/data'
     else:  # Parent. Wait for child and finish the work.
-      _, error = os.waitpid(pid, 0)
-      if error:
+      _, error_and_reason = os.waitpid(pid, 0)
+      if error_and_reason:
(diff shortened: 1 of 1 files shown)


sphynx/python/util.py
```diff
@@ -177,8 +177,9 @@ def mount(src, dst):
       os.chdir('/')
       self.datadir = '/data'
     else:  # Parent. Wait for child and finish the work.
-      _, error = os.waitpid(pid, 0)
-      if error:
+      _, error_and_reason = os.waitpid(pid, 0)
+      if error_and_reason:
+        error = error_and_reason // 256
```
xandrew-lynx, line 182: It must have seemed like a good idea 60 years ago... :)

xandrew-lynx (approved):

PR #141, opened 2021-02-11
Ditch ordered mapping

darabos:
The idea (from @xandrew-lynx) being that `MappingToOrdered` takes up a lot of memory. The tests seem to be passing locally. I haven't measured the impact on memory use yet. I also haven't thought backward compatibility entirely through.

### DIFF
--- a/sphynx/lynxkite-sphynx/vertex_set_intersection.go
+++ b/sphynx/lynxkite-sphynx/vertex_set_intersection.go
@@ -4,49 +4,21 @@ package main
 
 import (
 	"fmt"
-	"sort"
-	"sync"
 )
 
 type MergeVertexEntry struct {
 	id    int64
 	count int
 }
-type MergeVertexEntrySlice []MergeVertexEntry
-
-func (a MergeVertexEntrySlice) Len() int {
-	return len(a)
-}
-func (a MergeVertexEntrySlice) Less(i, j int) bool {
-	return a[i].id < a[j].id
-}
-func (a MergeVertexEntrySlice) Swap(i, j int) {
-	a[i], a[j] = a[j], a[i]
-}
-
-type Int64Slice []int64
-
-func (a Int64Slice) Len() int {
-	return len(a)
-}
-func (a Int64Slice) Less(i, j int) bool {
-	return a[i] < a[j]
-}
-func (a Int64Slice) Swap(i, j int) {
-	a[i], a[j] = a[j], a[i]
-}
 
 func doVertexSetIntersection(vertexSets []*VertexSet) (intersection *VertexSet, firstEmbedding *EdgeBundle) {
-	mergeVertices := make(MergeVertexEntrySlice, len(vertexSets[0].MappingToUnordered))
+	mergeVertices := make([]MergeVertexEntry, len(vertexSets[0].MappingToUnordered))
 	vs0 := vertexSets[0]
 	for idx, id := range vs0.MappingToUnordered {
 		mergeVertices[idx].id = id
 	}
-	sort.Sort(mergeVertices)
 	for i := 1; i < len(vertexSets); i++ {
-		w := make([]int64, len(vertexSets[i].MappingToUnordered))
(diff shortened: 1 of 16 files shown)


darabos:
I mean they passed before I added the assertion... 😅

darabos:
I did some testing.
image

The red is `master`, blue is this PR. The memory use decreased a bit, but the runtime increased a bit more. It's just 4 runs each with 8 different random graphs with 10,000,000 vertices and 30,000,000 edges. We compute shortest path on it. (This was one of the benchmark problems I used on the 256 GB test.)

sphynx/lynxkite-sphynx/vertex_set_intersection.go
```diff
@@ -4,49 +4,21 @@ package main
...
-func (a Int64Slice) Less(i, j int) bool {
-	return a[i] < a[j]
-}
-func (a Int64Slice) Swap(i, j int) {
-	a[i], a[j] = a[j], a[i]
-}
 
 func doVertexSetIntersection(vertexSets []*VertexSet) (intersection *VertexSet, firstEmbedding *EdgeBundle) {
-	mergeVertices := make(MergeVertexEntrySlice, len(vertexSets[0].MappingToUnordered))
+	mergeVertices := make([]MergeVertexEntry, len(vertexSets[0].MappingToUnordered))
```
xandrew-lynx, line 15: We could avoid this map as well via a multi-merge in a single nested loop. But no need to it in this PR.

xandrew-lynx (commented):
The code looks great to me!

Unfortunately, I guess. Given the benchmark it would have been a better news if I spotted something terrible in this... :)

I can kinda, reluctantly, accept that this can be slower, although even that I find unexpected. But how come we don't see a bigger memory impact? How did you measure memory usage?

darabos:
Thanks for the comments! (Hi Gabor!)

> I can kinda, reluctantly, accept that this can be slower, although even that I find unexpected. But how come we don't see a bigger memory impact? How did you measure memory usage?

I restarted Sphynx (but not LynxKite). Then I changed the random seed and waited for the average shortest path to get computed in this workspace:
image

Then I looked at "RES" in `top`. You can see the data I collected in the same benchmarking spreadsheet I used last week in the ["no mapping"](https://docs.google.com/spreadsheets/d/19u1zVM1KwJZ_VFsI7PvkLY7wJQNsUItBAHGbR4eiCd0/edit#gid=342664980) sheet.

I've set up `pprof` now. (Click to enlarge.)

|Before|After|
|---|---|
|profile002|profile004|

While this profile is talking about 1 GB, `top` shows 6.4 GB RES. I guess that's just fallout from memory management for GC.

olahg:
Awesome! I don't pretend I understand all of this, but here's some remarks:

- If it's just the memory usage you're worried about, maybe you could just use the ordered mapping when you need it, and then throw it away without caching it.
- The upper 32 bits of the original ids are random, but the lower 32 bits are totally predictable. Maybe it could be somehow utilized to save more memory.
- Sorting in Go is slow. I remember experimenting with it which working an the ranking attribute operation, and a c++ version was about twice as fast as the Go version. Maybe you could just sort lazily, only those vertex sets that really need it.
- You could sort considerably faster if you implemented your own non-interface-based custom sort. The interface-based solution calls a function for every swap and every comparison, which do their own bounds checking for every access.

darabos:
Thanks for the comments, Gabor!

I checked with `pprof` and indeed sorting is a significant part of it:
image

But a large part of sorting is the reflection! (In the central area.)

I switched to a fast concurrent sort library (https://github.com/jfcg/sorty) and avoided reflection during sorting. It's much better:

image

I can barely see sorting on this chart now. (There's a bit on the very left.)

This is reflected in the overall speed. The slight slowdown at least is gone.

image

Is this good to merge? Am I missing any compatibility issues?

olahg (approved):
LGTM

xandrew-lynx (approved):
Thanks, looks great!

PR #160, opened 2021-03-31
Table support in "Compute in Python" box

### LINKED ISSUE #72: Table support in Compute in Python
This would be the simplest thing. Just a DataFrame in and a DataFrame out.

It could work around the limitation that the box cannot change the graph structure. Just input the old edge list and output a new edge list. Maybe there's a better solution for this. But writing complex table manipulation code in Python would still be useful.

darabos:
Resolves #72.

I've been intensively using it for a few days now and it seems fine. I have a local hack with a TODO:

```
-      case t: TableData => t.df.write.parquet(getGUIDPath(e))
+      case t: TableData =>
+        // TODO: Instead of overwrite fix "has"?
+        t.df.write.mode(org.apache.spark.sql.SaveMode.Overwrite).parquet(getGUIDPath(e))
```

I'll sort that out and write tests and docs if the basic approach passes review!

Sure it would be nicer with multiple inputs, but I don't see a quick solution for that. You can combine stuff in SQL for now, before/after Python boxes.

Currently for graph inputs you output graphs and for table inputs you output tables. But maybe that doesn't have to be set in stone? If could output a table if we see `df.x` outputs and a graph if we see `vs.x`. But this may require another two backend operations (graph&rarr;table and table&rarr;graph), so I'm not eager to do it.


darabos:
I've fixed the hack, made it a bit more convenient to use and wrote a test. Documentation is still missing.

darabos:
I'm going to write documentation and merge this if you don't stop me, @xandrew-lynx. 😄

PR #175, opened 2021-04-29
[WIP] Add SpanningEdgeCentrality

darabos:
For @hampet97's thesis work. And it's nice anyway.

image

TODO: Test and docs.

For vertex centrality we have a single box. Do you think we should have a single box for edge centrality or separate boxes? With a single algorithm it feels weird to have a drop-down, but maybe better for the future?


darabos:
On further testing, there's something wrong with this algorithm. I get crashes that look like memory corruption on graphs a little bigger:

```
lynxkite-sphynx: malloc.c:4028: _int_malloc: Assertion `(unsigned long) (size) >= (unsigned long) (nb)' failed.
```

The forest fire score also works for Peter, so let's drop this for now.

PR #187, opened 2021-06-11
Use Scalafmt instead of Scalariform

### LINKED ISSUE #180: Switch from Scalariform to Scalafmt
Scalariform is not maintained. After #178 it says on every build that it reformatted 7 files, but I don't know which 7 and why they are always reformatted.

[Scalafmt](https://scalameta.org/scalafmt/) looks like a great replacement! I've tried the switch and it's very easy. It makes a lot of whitespace changes though, so I won't do it in #178.

Reading its docs made me realize that Scala 2.12 has **trailing commas!** I can't wait to use them. ❤️

darabos:
Resolves #180. I've reviewed all the diffs and made small tweaks here and there. The changes fall into three categories:
- Diffs that improved formatting. I love the trailing commas! Changing `1l` to `1L` is also great.
- Diffs I don't care about. This is the vast majority.
- Function calls with a lot of short parameters each on a separate line. I've asked about this in https://stackoverflow.com/questions/67936058/keep-multiple-parameters-on-one-line-in-scalafmt.

I've committed the formatting changes as a fictional `scalafmt` user to try to not skew contributor metrics. I don't like large diffs, but I like dead dependencies even less. I think we'll be fine with this.


darabos:
Looks like a new version of something (mypy?) requires a new dependency.

```bash
+ python3 -m mypy src/lynx --ignore-missing-import
src/lynx/kite.py:35: error: Library stubs not installed for "requests" (or incompatible with Python 3.7)
src/lynx/kite.py:35: note: Hint: "python3 -m pip install types-requests"
src/lynx/kite.py:35: note: (or run "mypy --install-types" to install all missing stub packages)
src/lynx/kite.py:35: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

I've tried adding `--install-types` but it pops an interactive prompt. I'm not sure how the dependencies end up installed in the GitHub environment... Hopefully it's from `setup.py`.

darabos:
This is 99% whitespace changes. I think it's not important to review.

PR #188, opened 2021-06-25
Cache surprisingly slow findVariables calls

darabos:
Other functions in this file that run code are already cached, but I guess this was not slow enough. But I now have a workspace where it would take a couple seconds to just open the parameters of a Derive box. This fixed it.

### DIFF
--- a/app/com/lynxanalytics/sandbox/ScalaScript.scala
+++ b/app/com/lynxanalytics/sandbox/ScalaScript.scala
@@ -286,17 +286,22 @@ object ScalaScript {
     }
   }
 
+  private val findVariablesCache = new SoftHashMap[String, Set[String]]()                                             
+  
   def findVariables(code: String): Set[String] = {
     import scala.reflect.internal.util.ScriptSourceFile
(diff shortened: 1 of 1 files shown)


app/com/lynxanalytics/sandbox/ScalaScript.scala
```diff
@@ -286,17 +286,22 @@ object ScalaScript {
     }
   }
 
+  private val findVariablesCache = new SoftHashMap[String, Set[String]]()
```
olahg, line 289: Is this thread-safe?

olahg: Sorry for the stupid question. It is thread-safe.

olahg (approved):
LGTM

app/com/lynxanalytics/sandbox/ScalaScript.scala
darabos: > Sorry for the stupid question. It is thread-safe.

Oh, thanks for checking! I just copied it from the other cached things in this file.

PR #194, opened 2021-08-13
Upgrade to Spark 3.1.2, link NetworKit statically, and more

darabos:
I still have to check the GitHub Actions build. It's probably not working at this point.


darabos:
Look at that! The build is finally fixed!

PR #197, opened 2021-08-18
Built-in box for ROC curve

### LINKED ISSUE #181: New built-in: ROC curve
I can never remember where I can copy it from. 😅

darabos:
Resolves #181.

In action:

image

And inside:

image


built-ins/draw-ROC-curve
```diff
@@ -0,0 +1,155 @@
...
+        / (select sum(label) from input)
+        as tpr,
+
+      sum(1 - label) over (
+        order by score desc rows between
+        unbounded preceding and current row)
+        / (select sum(1 - label) from input)
+        as fpr
+
+      from input
```
darabos, line 114: @erbenpeter when you are not on vacation I'd appreciate if you could take a look at this SQL query for calculating the curve.

built-ins/draw-ROC-curve
```diff
@@ -0,0 +1,155 @@
...
+  inputs:
+    input:
+      boxId: SQL1_5
+      id: table
+  operationId: SQL1
+  parameters:
+    persist: 'no'
+    sql: |
+      select sum(1 - fpr) / count(1) as AUC
+      from input where label == 1
```
darabos, line 129: Plus this one for calculating the AUC. I made up both queries myself instead of copying them from somewhere, so I'm not terribly confident in them. 😅  Thanks!

darabos:
Everyone's on vacation but I want to prepare a release candidate. I'm merging this while hoping for a future review.

erbenpeter:
I'm back from vacation, so now I'll take a look.

built-ins/draw-ROC-curve
```diff
@@ -0,0 +1,155 @@
...
+      boxId: SQL1_4
+      id: table
+  operationId: SQL1
+  parameters:
+    persist: 'no'
+    sql: |-
+      select
+      label, score,
+      sum(label) over (
+        order by score desc rows between
```
erbenpeter, line 103: I only found a small issue here: If it's possible that the classifier assigns the same score for multiple items, for that specific score we will get a lot of different `(fpr, tpr)` pairs. Only the last one (order by score desc) is really meaningful.

On the chart it can create weird blobs I think.

One possible solution is to add a second sql to pick the last value from each score-group, but it only works if we can assume that the order of the records remains the same between the first and the second sql. Or we can pre-aggregate by score `(num_1, num_0)` and the then your SQL will work even for repeated score values.

built-ins/draw-ROC-curve
```diff
@@ -0,0 +1,155 @@
...
+- id: SQL1_6
+  inputs:
+    input:
+      boxId: SQL1_5
+      id: table
+  operationId: SQL1
+  parameters:
+    persist: 'no'
+    sql: |
+      select sum(1 - fpr) / count(1) as AUC
```
erbenpeter, line 128: It's a tricky implementation (compared to the first 10 I've found googling) but it also assumes that scores are unique (which I think is not realistic: if two customers have the same ingredients the (deterministic) classifier needs to assign the same score to them, but it's possible that their labels are different in reality.)

Mathematically speaking this formula only works when the segments of the ROC curve are all vertical or horizontal.

erbenpeter (commented):
Thanks, nice and clean!

I  almost like it. My issue is with the possibility of items have the same score but different labels (which I think is possible). See my more detailed comments inline.

erbenpeter (commented):
Thanks, nice and clean!

I  almost like it. My issue is with the possibility of items have the same score but different labels (which I think is possible). See my more detailed comments inline.

built-ins/draw-ROC-curve
erbenpeter: To phrase my problem in a different way: if it's possible to have repeated scores with different labels, your curve is not defined, its shape (and consequently the AUC value) )depends on the order of the data points inside a group with the same score.

built-ins/draw-ROC-curve
```diff
@@ -0,0 +1,155 @@
...
+    summary: Rename and filter
+  parametricParameters:
+    sql: |-
+      select
+        ${`true label`} as label,
+        ${`predicted score`} as score
+      from vertices
+      where isnotnull(${`true label`})
+        and isnotnull(${`predicted score`})
+      limit ${`sample size`}
```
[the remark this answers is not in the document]
darabos: > Can we regard this a a true random sample because the rows are in random order in the query?

Yes. The input is a graph where this usually holds.

app/com/lynxanalytics/biggraph/serving/JsonServer.scala
```diff
@@ -392,8 +392,7 @@ class ProductionJsonServer @javax.inject.Inject() (
 
   def downloadCSV = asyncAction(parse.anyContent) { (user: User, r: mvc.Request[mvc.AnyContent]) =>
     val request = parseJson[GetTableOutputRequest](user, r)
-    implicit val metaManager = workspaceController.metaManager
-    val table = workspaceController.getOutput(user, request.id).table
+    val table = workspaceController.metaManager.table(java.util.UUID.fromString(request.id))
```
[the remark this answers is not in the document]
darabos: > This part I don't understand even after reading the long commit command. How is it related to the newly added built-in box?

The plot refers to the table GUID and the frontend sends this request to get the data. The box output state for a table also has the table's GUID. So you can access the same table either by looking up the GUID as a box output state ID and then taking the table from it (the old code) or looking up the GUID as a table (the new code).

Box output states are not persisted. We assume you only want to look at a box output that we have returned in this run. So if you restart LynxKite after creating a plot, and then look at the plot without looking at the box that generated it, you get an error. This is an edge case I didn't consider originally. You don't typically look at box outputs when not looking at the box. Except this happens with custom boxes!

built-ins/draw-ROC-curve
darabos: > I only found a small issue here: If it's possible that the classifier assigns the same score for multiple items, for that specific score we will get a lot of different `(fpr, tpr)` pairs. Only the last one (order by score desc) is really meaningful.
> 
> On the chart it can create weird blobs I think.

Nope. Looks fine:

image

built-ins/draw-ROC-curve
darabos: > To phrase my problem in a different way: if it's possible to have repeated scores with different labels, your curve is not defined, its shape (and consequently the AUC value) )depends on the order of the data points inside a group with the same score.

You need to have them in a random order, which is the case here. So you get a nearly diagonal line and an AUC close to 0.5 in the case of the above screenshot. (Random 0/1 label and random 0/1 prediction.)

darabos: > Mathematically speaking this formula only works when the segments of the ROC curve are all vertical or horizontal.

Then I have great news! The segments of the ROC curve are all vertical or horizontal.

darabos:
> I almost like it. My issue is with the possibility of items have the same score but different labels (which I think is possible). See my more detailed comments inline.

I've checked how scikit-learn does the ROC curve when scores are repeated and it does it like you say. 😓  Sorry I was confidently wrong. I've sent a fix in #202. Thanks!

PR #202, opened 2021-09-10
Use sklearn for ROC curve

darabos:
Thanks @erbenpeter for the comments in #197. I added the sampling to avoid pushing a million points to the plot. But since it's sampled anyway, we can also make the ROC curve in Python. It's one line with scikit-learn and it properly handles the case where the same score is returned multiple times.

Before/after for random 0/1 scores on a balanced dataset:
image


erbenpeter (commented):
Look nice.

PR #205, opened 2021-09-24
Use a fixed version of protoc and upgrade related dependencies to latest

darabos:


sphynx/proto_compile.sh
```diff
@@ -28,14 +28,14 @@ if [[ ! -f $REPO/.build/$GRPC_JAVA ]]; then
   wget -nv -P .build https://repo1.maven.org/maven2/io/grpc/protoc-gen-grpc-java/$GRPC_JAVA_VERSION/$GRPC_JAVA
   chmod +x $REPO/.build/$GRPC_JAVA
 fi
-protoc --plugin=protoc-gen-grpc-java=$REPO/.build/$GRPC_JAVA --grpc-java_out=../app \
+protoc/bin/protoc --plugin=protoc-gen-grpc-java=$REPO/.build/$GRPC_JAVA --grpc-java_out=../app \
```
darabos, line 31: Maybe I'm missing something. But it looks like we were downloading protoc 3.10.1 and then using the system protoc instead.

PR #209, opened 2021-10-07
Use an environment instead of manually installing packages

darabos:
Wow, Conda can install everything! The GitHub Actions now run without Docker.

### DIFF
--- a/test/com/lynxanalytics/biggraph/graph_operations/DeriveScalaTest.scala
+++ b/test/com/lynxanalytics/biggraph/graph_operations/DeriveScalaTest.scala
@@ -289,8 +289,7 @@ class DeriveScalaTest extends AnyFunSuite with TestGraphOp {
     val e = intercept[org.apache.spark.SparkException] {
       derived.rdd.collect
     }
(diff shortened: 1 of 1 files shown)


darabos:
The backend test failures are:

```
[info] - example graph - security manager *** FAILED *** (2 seconds, 858 milliseconds)

[info]   Caused by: java.security.AccessControlException: access denied ("java.io.FilePermission" "abc" "read")
```

```
[info] - user defined functions - geodistance *** FAILED *** (14 seconds, 34 milliseconds)
[info]   ArrayBuffer(ArrayBuffer(7023993.307994275), ArrayBuffer(7023993.307994275), ArrayBuffer(1.5340398666732997E7), ArrayBuffer(9507129.781908857)) did not equal List(List(7023993.307994274), List(7023993.307994274), List(1.5340398666732997E7), List(9507129.781908857)) (SQLTest.scala:412)
```

Both of these look like they may be caused by minor formatting differences, e.g. from using a different Java version. I'll try to just make the tests more robust.

The Sphynx test is broken due to #206. I'll fix that in a separate PR.

Issue #151, opened 2021-03-05
Remove "limit 10" from the default SQL queries


xandrew-lynx:
The result is limited on display anyways. I delete it every single time and it's annoying. Any drawback of removing these?

xandrew-lynx:
(Except when I forget to remove it and scratch my head on why something is weird downstream... :/ )

darabos:
Ah, yes please! The counter argument was (as I recall) that "persist: yes" is also a default. So even though in `getTableOutput` we only fetch the first 10 lines, the backend may still end up computing and storing a lot of data unnecessarily.

How about we set "persist: no" as the default? I think the counter argument for that was that Spark at the time didn't handle long chains well. But I'm just upgrading to Spark 3. Hopefully it's a solved problem by now. A longer chain even allows for better optimization. (E.g. maybe you don't need to compute some intermediate data if you don't need it in the end.) And also we are not building long SQL chains all the time. You can turn on "persist" when needed.

xandrew-lynx:
Sounds good to me. Especially as persisting also messes up ordering when the user adds "order by". :/

darabos:
Done in #220. Forgot to close it.

PR #220, opened 2022-02-02
Faster workspace execution

darabos:
I'm addressing multiple issues here:
- Persistent outputs are slower because they write to disk. It's just a bad default.
- The default LIMIT is annoying and no longer necessary if we don't persist the table. (#151)
- Each SQL box created 6 VertexToEdgeAttribute operations per input vertex attribute. With 100 attributes and 20 SQL boxes this was 12,000 operations on every `setAndGetWorkspace` request. That takes around 3 seconds. That's long enough that you start clicking on more things, creating an increasing backlog of requests. In my test I got 20–30 second response times easily.

Screenshot 2022-02-01 at 19 31 43

For the third issue 4 of the 6 VertexToEdgeAttribute operations were unnecessarily caused by the `defaultTableName` code. That's easy to fix. But the other two are unavoidable if we execute the box.

So I've added a "BoxCache" to avoid unnecessary box executions. This is a huge improvement. Even if you only moved a box, we used to execute all the boxes in the workspace. But with the cache we don't execute anything and it's super fast. When you change something, we only execute that one box.

I went with this universal solution rather than try to improve the SQL box code, because it's not the only box that scales linearly with the number of attributes. E.g. a filter box will pull over all attributes. It's a bit faster than the SQL box, but I still saw 1–2 second latencies with 100 attributes and 20 filter boxes. The cache fixes this for all boxes.

Do you see any potential issues with it?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Please add changes to "master", preferably ordered by their significance. (Most 
 ### master
 
 - Upgraded to PyTorch Geometric (PyG) 2.0.1. [#206](https://github.com/lynxkite/lynxkite/pull/206)
+- The workspace interface is much faster now. [#220](https://github.com/lynxkite/lynxkite/pull/220)
 
 ### 4.3.0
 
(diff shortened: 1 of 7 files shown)


darabos:
Thanks!

PR #222, opened 2022-02-18
LynxKite 4.4.0 release

darabos:
The half-duplicated Conda environment is not great. But we also can't put the whole dev environment in the Docker image. Let me know if you have a better idea.

Also included are some fixes for warnings about deprecations.


darabos:
This seems to work okay. I'll take a look at the state of the tests after merging it and hopefully finally publish the release.

darabos:
Sorry, this ran longer than intended. On the positive side, we've been running LynxKite 4.4.0 in production since March. I just need to upstream a few last-minute changes and I can release the open-source 4.4.0 too.

PR #234, opened 2022-05-23
Very last changes for 4.4.0

darabos:
Added a few missed items to CHANGELOG and dependencies. If tests pass, this will be it!


darabos:
Okay, I've seen each test pass, though not at the same time. The frontend test is still unreliable inside GitHub Actions. 😞  Let's release this!

PR #240, opened 2022-05-24
Python API for interfacing with PySpark DataFrames

darabos:
It's just a few lines, but it makes the Spark–LynxKite integration much more friendly. From Spark to LynxKite:

```python
from pyspark.sql import Row, SparkSession
spark = SparkSession.builder.getOrCreate()
people_df = spark.createDataFrame([
    Row(name='Adam', age=23),
    Row(name='Eve', age=34),
    Row(name='Bob', age=45),
])
relationships_df = spark.createDataFrame([
    Row(a='Adam', b='Eve', weight=3),
    Row(a='Bob', b='Eve', weight=4),
    Row(a='Eve', b='Adam', weight=5),
])
g = lk.from_spark(people_df).useTableAsVertices()
g = lk.useTableAsEdges(
    g,
    lk.from_spark(relationships_df),
    attr='name', src='a', dst='b')
```

From LynxKite to Spark:
```python
t = g.sql('select count(*) from vertices')
t.spark(spark).show()
```

PySpark and LynxKite in this case still use separate Spark applications. But perhaps they are both dynamically scaled? The important thing is that this is a fully distributed data exchange between the two systems.

### DIFF
--- a/conda-env.yml
+++ b/conda-env.yml
@@ -40,3 +40,5 @@ dependencies:
 - requests
 - typed-ast
 - types-requests
+# For Python tests.
+- pyspark==3.2.1
--- a/python/remote_api/src/lynx/kite.py
+++ b/python/remote_api/src/lynx/kite.py
@@ -984,7 +984,7 @@ class State:
   def get_progress(self):
     '''Returns progress info about the state.'''
     state_id = self._get_state_id()
-    progress = lk._ask('/ajax/long-poll', dict(syncedUntil=0, stateIds=[state_id]))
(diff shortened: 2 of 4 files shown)


darabos:
Test failure:
```
FAIL: test_pyspark (test_external_computation.TestExternalComputation)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/lynxkite/lynxkite/python/remote_api/tests/test_external_computation.py", line 63, in test_pyspark
    self.assertTrue(actual.equals(expected))
AssertionError: False is not true
```

PR #241, opened 2022-05-25
Run algorithms on a GPU with NVIDIA RAPIDS

darabos:
Just PageRank for now, but I'll add more in a bit.

I wanted to add some benchmark numbers too, but PageRank cannot be directly compared. CuGraph doesn't run for a set number of iterations but rather until convergence according to a tolerance parameter.


darabos:
I've measured betweenness centrality on a 6,000,000-edge graph with 1000 samples. NetworKit is 11.4 minutes, CuGraph is 1.3 minutes.

darabos:
This is ready for review!

I recognize that using the NetworKit backend operations for CUDA leads to confusing naming. But the generality of the "NetworKit" operation signatures is fantastic. We can add new operations this way without writing any boilerplate Scala code. I think we have a tradition of preferring backward compatibility over good naming. 😅 

CuGraph also has a number of algorithms that we don't have in any boxes. It's tempting to add them, but we have to save something for future versions too.

I'm a bit reluctant to add GPU-only boxes that go away without `KITE_ENABLE_CUDA`. Or is that okay?

darabos:
Merging this for now. But comments are welcome anyway! Especially let me know if there are cuGraph algorithms I should add as new boxes.

PR #244, opened 2022-06-08
[WIP] Add Jaccard similarity box

darabos:
When writing a test I discovered we're getting some broken values out of this. I've filed https://github.com/networkit/networkit/issues/953 with NetworKit. But I'm afraid we can't put it in LynxKite 5.0.


darabos:
The response from the NetworKit side was that this operation only supports graphs without parallel edges. I guess we can work with that. But we have too much going on now. Let's come back to this later.

PR #257, opened 2022-07-14
Allow edge attributes to be used as weight in 'weighted aggregate on neighbours'

### LINKED ISSUE #230: Option for weighing by edges at the "weighted aggregate on neighbours function"
Dear LK Developers, 

It would be nice to add the option of weighting the node values by the edges at the weighted aggregate on neighbors function. I have a workaround (create a vertex feature from the edge-weight and use that as a weight), but it would be more comfortable to have the option as well.

Thank you.

Best Regards,
Marton

lacca0:
Resolves #230


darabos (commented):
Almost done! 👍

darabos:
Oh, and please add a test! There is hopefully already a test for this box, so just add a new entry there that tests it with an edge attribute. Thanks!

darabos:
Sorry, I thought this was merged already. Do you need more review from me? I think only the tests are missing. While there is no urgency for this, I think it's better to not put this off too long. Or we will just forget what the whole thing was about by the time you get to it. 😄

PR #263, opened 2022-07-28
[WIP] Add aggregation tests

lacca0:
Not happy with the type conversion in AggregateFromSegmentationOperationTest.scala yet

### DIFF
--- a/test/com/lynxanalytics/biggraph/frontend_operations/AggregateOnNeighborsTest.scala
+++ b/test/com/lynxanalytics/biggraph/frontend_operations/AggregateOnNeighborsTest.scala
@@ -51,3 +51,24 @@ class AggregateOnNeighborsTest extends OperationsTestBase {
       2 -> Vector(41, -74, 48, 19)))
   }
 }
+
+class WeightedAggregateOnNeighborsTest extends OperationsTestBase {
+  test("all aggregators") {
+    def agg[T](attribute: String, aggregator: String, weight: String): Map[Long, T] = {
+      val p = box("Create example graph")
+        .box(
+          "Weighted aggregate on neighbors",
+          Map(
+            "prefix" -> "",
+            "direction" -> "all edges",
+            "weight" -> weight,
+            ("aggregate_" + attribute) -> aggregator))
+        .project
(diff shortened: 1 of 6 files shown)


darabos (approved):
Awesome, thanks!

GitHub shows a ⛔ sign at the end of your files to show that there's no newline at the end. Please add a newline. (Or run `scalafmtAll` in `sbt`. I think it will fix the files automatically.)

> Not happy with the type conversion in AggregateFromSegmentationOperationTest.scala yet

What's wrong with it? It seems okay to me.

Most of the tests directly access RDDs, like you do. But there is a newer alternative that you can see for example in `CreateGraphInPythonTest.scala`. You call `get()` on an attribute (or anything else) and get back a plain Scala structure. Maybe this is what you're looking for?

I like `get()` better, but the RDD-based tests work fine too so we didn't rewrite the existing tests with `get()`.

darabos:
The `get()` methods are defined in [GraphTestUtils.scala](https://github.com/lynxkite/lynxkite/blob/5.0.0/test/com/lynxanalytics/biggraph/graph_api/GraphTestUtils.scala#L103) if you want to see.

test/com/lynxanalytics/biggraph/frontend_operations/AggregateOnNeighborsTest.scala
```diff
@@ -51,3 +51,24 @@ class AggregateOnNeighborsTest extends OperationsTestBase {
...
+      val p = box("Create example graph")
+        .box(
+          "Weighted aggregate on neighbors",
+          Map(
+            "prefix" -> "",
+            "direction" -> "all edges",
+            "weight" -> weight,
+            ("aggregate_" + attribute) -> aggregator))
+        .project
+      get(p.vertexAttributes(attribute + "_" + aggregator + "_by_" + weight)).asInstanceOf[Map[Long, T]]
```
darabos, line 67: I think this should also work:
```suggestion
      get(p.vertexAttributes(attribute + "_" + aggregator + "_by_" + weight).runtimeSafeCast[T])
```

darabos (approved):

Issue #265, opened 2022-08-19
Deprecated functions errors when compile against Java 17


lacca0:
When we compile lynxkite against Java 17 we get method/class deprecated errors - the whole SecurityManager concept is removed from Java 17: https://openjdk.org/jeps/411. We must find a proper substitution.

```
[error] /home/lynx/Documents/lynxkite/app/com/lynxanalytics/sandbox/ScalaScript.scala:28:10: method setSecurityManager in class System is deprecated
[error]   System.setSecurityManager(restrictedSecurityManager)
[error]          ^
[error] /home/lynx/Documents/lynxkite/app/com/lynxanalytics/sandbox/ScalaScript.scala:32:42: class SecurityManager in package lang is deprecated
[error] class ScalaScriptSecurityManager extends SecurityManager {
[error]                                          ^
[error] /home/lynx/Documents/lynxkite/app/com/lynxanalytics/sandbox/ScalaScript.scala:43:31: class AccessControlException in package security is deprecated
[error]       throw new java.security.AccessControlException("Access error: GET_THREAD_GROUP")
[error]                               ^
[error] /home/lynx/Documents/lynxkite/app/com/lynxanalytics/sandbox/ScalaScript.scala:104:31: class AccessControlException in package security is deprecated
[error]       throw new java.security.AccessControlException(s"Illegal package access: $s")
[error]                               ^
```

darabos:
[JEP 411: Missing use-case: user functions in an RDBMS](https://marc.info/?t=162216600200001&r=1&w=2) is a nice thread by @jcflack about a situation similar to ours. https://inside.java/2021/04/23/security-and-sandboxing-post-securitymanager/ is an article from one of the proponents of JEP 411 about the topic. Neither has a practical recommendation.

PL/Java has decided to stick with SecurityManager and tells users to just ignore the warnings. (https://github.com/tada/pljava/wiki/JEP-411) We can consider doing the same. I see some [pushback](https://jep411.github.io/) about JEP 411. Maybe if we close our eyes it will go away. With a Scala 2.12.13 upgrade we may be able to suppress this warning. (https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html)

The alternative is to say goodbye to SecurityManager. Remove all security restrictions on Scala code, disable the "Derive attribute" boxes by default and provide a setting to turn them on. Just like how we have `KITE_ALLOW_PYTHON` instead of sandboxing Python.

PR #266, opened 2022-08-19
Force conda to use java11

lacca0:
Temporary workaround for #265


darabos (approved):
Thanks! Also thanks to @tuckging for figuring this out!

darabos:
The test failure is due to the issue fixed in #261. I'll just merge everything and we can fix any real issues in follow up PRs.

PR #269, opened 2022-08-24
Build a fat jar that works with spark-submit

darabos:
This is a substantial change to how LynxKite runs. We pack LynxKite (including Sphynx) into a single jar file that can be started with `spark-submit`. This makes it way easier to deploy in a Hadoop environment. We're open-sourcing this from a private repo (sorry) where it has already proven useful in two enterprise environments.

PR #273, opened 2022-08-30
Prepare changelog for LynxKite 5.1.0

darabos:
Let me know if I missed anything noteworthy.


darabos:
I think there will be one more big change the release, to update the Docker image for #269. But that's not noteworthy on its own.

lacca0 (approved):

PR #277, opened 2022-09-06
Add fat jar to Scala interpreter classpath

darabos:
I think `embeddedDefaults` was supposed to do something similar, but it doesn't work. The classpath has to be set before the interpreter is created. See https://stackoverflow.com/questions/51723400/issues-with-scala-scriptengine-inside-spark-submit-application.


darabos:
At last! All tests are green!

image

Last I checked even the Docker image was working. Please review these quickly before the tests break again! Thanks!

darabos:
I'm still looking forward to the review, but I'll merge this too so we have something 100% working on main.

PR #280, opened 2022-09-12
Fix paragraphs overlapping in comments

darabos:
I'm not sure when this happened — I just noticed it on try.lynxkite.com. I don't think we have changed this recently, so maybe it's a change in Chrome?

|before fix|after fix|
|---|---|
|image|image|

### DIFF
--- a/web/app/styles/workspace.css
+++ b/web/app/styles/workspace.css
@@ -300,6 +300,10 @@ op-editor > :first-child {
   padding-top: 10px;
 }
 
+.workspace-comment {
+  display: block;
+}
+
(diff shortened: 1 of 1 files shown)


darabos:
The frontend test failure is on the same test as I saw failing earlier. We should check into it when we have time.

```
1) regression data imported as vertices can be reached
  Message:
    Failed: No element found using locator: By(css selector, #side-left)
  Stack:
    NoSuchElementError: No element found using locator: By(css selector, #side-left)
```

lacca0 (approved):

PR #286, opened 2022-09-23
Comments for a few tricky places

### LINKED ISSUE #260: Add documentation for LynxKite core mechanisms
While onboarding @lacca0 I explained a lot of what's happening in `MetaGraph.scala` and other core files. But we really should have more of this explained in the form of comments. This code almost never changes, so maintaining the comments will not be a big deal. On the other hand this is some really tricky code.

darabos:
Resolves #260. To up the stakes, I've also added some code changes. 😈 

Any other dark areas I should try to illuminate?

### DIFF
--- a/app/com/lynxanalytics/biggraph/graph_api/MetaGraph.scala
+++ b/app/com/lynxanalytics/biggraph/graph_api/MetaGraph.scala
@@ -5,9 +5,14 @@
 // the inputs and outputs of the operation instances. The operation instance is
 // an operation that is bound to a set of inputs.
 //
+// On this "meta" level, there is no data. It is just about identifying the entities.
+// The entities are identified by UUIDs in storage. These UUIDs (or guids) are a hash of
+// the whole tree of the metagraph leading up to the entity. This is how we track what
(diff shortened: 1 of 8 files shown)


app/com/lynxanalytics/biggraph/graph_api/MetaGraph.scala
```diff
@@ -55,23 +61,6 @@ sealed trait MetaGraphEntity extends Serializable {
     .sliding(2, 2).map(Integer.parseInt(_, 16).toByte).toSeq
 }
 
-case class StringStruct(name: String, contents: SortedMap[String, StringStruct] = SortedMap()) {
```
darabos, line 58: Okay, so this was still used by a test. I've changed the test, but it's less thorough now. I think that's an okay trade-off. I would rather have a less precise test for a rarely used feature than having to maintain this complex code with reflection and everything.

It would also be possible to rewrite the test a bit more so it keeps the exact same check. That would be a bit more work though.

lacca0 (approved):

PR #291, opened 2022-09-29
Allow starting and stopping LynxKite from Scala

darabos:
The idea is that you have a JVM which already has a Spark session. You want to run LynxKite in this session. And you want to use it from Python too while it's running. This is a common situation in a Databricks notebook, which allows mixing Scala and Python cells.

Instead of a `kiterc` you can set environment variables or provide overrides like this:

```scala
com.lynxanalytics.lynxkite.Environment.set(
  "KITE_ENABLE_CUDA" -> "yes",
  "KITE_CONFIGURE_SPARK" -> "no",
  "KITE_META_DIR" -> "/home/darabos/kite/meta",
  "KITE_DATA_DIR" -> "file:/home/darabos/kite/data",
  "KITE_ALLOW_PYTHON" -> "yes",
  "KITE_ALLOW_NON_PREFIXED_PATHS" -> "true",
  "SPHYNX_HOST" -> "localhost",
  "SPHYNX_PORT" -> "5551",
  "ORDERED_SPHYNX_DATA_DIR" -> "/home/darabos/kite/sphynx/ordered",
  "UNORDERED_SPHYNX_DATA_DIR" -> "/home/darabos/kite/sphynx/unordered",
)
com.lynxanalytics.lynxkite.Main.start()
// ...
com.lynxanalytics.lynxkite.Main.stop()
```

All this wouldn't be too bad. But it's the first time we're really exposing the LynxKite package name. I wanted it to be `com.lynxanalytics.lynxkite` rather than `com.lynxanalytics.biggraph`. So there are a bit more diffs than strictly necessary.

But we should have renamed it already anyway! What's a "biggraph"? Nobody knows.


darabos:
Sorry, the last commit is not strictly relevant, but it fixes an issue I ran into testing this on Databricks. In a single-node cluster `sparkContext.isLocal` is `true`, so we can directly copy from local disk to distributed storage. But the default filesystem is still `dbfs:`, not `file:`. So I'm adding `file:` explicitly, rather than relying on the default.

With these changes it's working! https://gist.github.com/darabos/d8e107bd4bc437166e96700fee652588

lacca0 (approved):

darabos:
After thinking about it a bit more, it would take too many hacks to rename `biggraph` to `lynxkite` with compatibility.

There are still two options left:
- Go with `com.lynxanalytics.biggraph.LynxKite`. Because it's not `Main` any longer, it's pretty clear what it is, so go with the simplest option. Or...
- Keep everything under `biggraph`, except for what we want to expose as the public API. `LynxKite` and `Environment` go to the `lynxkite` package. This keeps compatibility safe and the public package name pretty. Plus it makes it easy to tell what classes are public.

I like both. With the second one I'm slightly worried about the evolution over time. Will we keep moving classes between the two packages as we expand the API? And I'm not even sure about the name. `com.lynxanalytics.lynxkite`? Or `com.lynxkite` to match the web site? With option one we avoid having this discussion now.

PR #294, opened 2022-10-12
Start and stop LynxKite from Python

darabos:
The more generally useful descendant of #291. The example works!

```python
import lynx.kite
from pyspark.sql import SparkSession
spark = lynx.kite.add_jar(SparkSession.builder).getOrCreate()
lk = lynx.kite.LynxKite(spark=spark)
print(lk.createExampleGraph().sql('select * from edges').df())
```

Next steps:
- [x] Figure out how to get the jar path. (Currently it's hardcoded. 😓)
- [ ] Try it on Databricks.
- [x] Automated test.
- [x] Docs. I'm not sure where to document this. Maybe just the changelog for now.
- [ ] Create Conda package with LynxKite + its Python API, like how Spark is bundled. (This will be a separate PR at some point.)
- [ ] Automatically start and stop when executing something. Currently it's just started at the start and never stopped. While this would be possible to do, I'm not sure it's necessary. Everything will be destroyed anyway when Spark is shut down. Stopping just LynxKite doesn't do much.
- [ ] Next up: start and stop LynxKite from R. (Just kidding! 😄)


darabos:
In light of this the renaming part of #291 really doesn't matter. Nobody will use the Scala start/stop API. But I merged it because it still had the dedicated start code, the environment overrides, and a bunch of bug fixes. Those are all useful.

python/remote_api/src/lynx/kite.py
```diff
@@ -854,6 +874,55 @@ def set_executors(self, count):
     return self._send('/remote/setExecutors', {'count': count})
 
 
+def get_free_ports(n):
```
lacca0, line 877: Please add a docstring. And make this private.

python/remote_api/src/lynx/kite.py
```diff
@@ -854,6 +874,55 @@ def set_executors(self, count):
...
+  servers = [socketserver.TCPServer(('localhost', 0), None) for _ in range(n)]
+  for s in servers:
+    s.__enter__()
+  ports = [s.server_address[1] for s in servers]
+  for s in servers:
+    s.__exit__()
+  return ports
+
+
+class ManagedLynxKite:
```
lacca0, line 887: Same here.

python/remote_api/src/lynx/kite.py
```diff
@@ -854,6 +874,55 @@ def set_executors(self, count):
...
+  def set_env(self, **kwargs):
+    jvm = self.spark._jvm
+    cfg = jvm.java.util.ArrayList()
+    for k, v in kwargs.items():
+      cfg.append(jvm.Tuple2(k, str(v)))
+    cfg = jvm.scala.collection.JavaConversions.asScalaBuffer(cfg)
+    jvm.com.lynxanalytics.biggraph.Environment.set(cfg)
+
+  def start(self):
+    self.spark._jvm.com.lynxanalytics.biggraph.LynxKite.start()
```
lacca0, line 920: When does this function return?

lacca0 (approved):

python/remote_api/src/lynx/kite.py
darabos: > Please add a docstring. And make this private.

Done.

python/remote_api/src/lynx/kite.py
darabos: > When does this function return?

It appears to return when LynxKite is ready to receive requests. I've added a docstring to make this official.

I have to admit I'm not sure what ensures this. In `LynxKite.scala` we call `play.core.server.ProdServerStart.start()`. My guess is it only returns when everything is ready.

The reason I think this works correctly is because the tests pass reliably.

Issue #293, opened 2022-10-11
Visualizations in Python and R


darabos:
Python and R have fantastic visualization libraries. Some of them output fancy interactive objects, but the majority of them output plain bitmaps. We should have a box where we output this image.

This would be a new type of output. (Not a graph/table/visualization/Vega plot.) In fact we could make this output able to contain arbitrary HTML, and serve up the image that way. That way we leave the door open for adding interactive charts and other gadgets later.

(We have long talked about automatically generating a detailed report on a graph. That report could also take advantage of this HTML output type. Just make sure we don't allow user-controlled HTML in multi-user environments.)

PR #297, opened 2022-10-27
Plotting from "Compute in Python"

darabos:
For #293.

Another step closer to turning LynxKite into Jupyter! 😄 

image


app/com/lynxanalytics/biggraph/controllers/Workspace.scala
```diff
@@ -398,7 +399,7 @@ object BoxOutputState {
     BoxOutputState(BoxOutputKind.Plot, Some(plot))
   }
 
-  def from(
+  def exportResult(
```
darabos, line 402: This couldn't stay `from()` because we have two outputs that are `Scalar[String]` — export results and HTMLs. Hence the renaming.

app/com/lynxanalytics/biggraph/frontend_operations/Operations.scala
```diff
@@ -739,4 +742,31 @@ object PythonUtilities {
...
     op(op.df, table).result.df
   }
+
+  def deriveHTML(
+      code: String,
+      inputs: Seq[String],
+      project: com.lynxanalytics.biggraph.controllers.ProjectEditor)(
+      implicit manager: MetaGraphManager): Scalar[String] = {
+    val api = Seq("vs", "es", "graph_attributes")
+    // Run the operation.
```
darabos, line 752: This is also mostly shared with `derive`, but I just copy-pasted it instead of abstracting it out.

app/com/lynxanalytics/biggraph/frontend_operations/WorkflowOperations.scala
```diff
@@ -609,8 +609,14 @@ class WorkflowOperations(env: SparkFreeEnvironment) extends ProjectOperations(en
         input.kind match {
           case BoxOutputKind.Project =>
             val project = projectInput("graph")
-            PythonUtilities.derive(params("code"), pythonInputs, pythonOutputs, project)
-            Map(context.box.output("graph") -> BoxOutputState.from(project))
+            if (params("outputs") == "plot") {
```
darabos, line 612: So you indicate that you want a plot by just declaring the output as `plot`. I think that's not too bad, since the normal output declarations will always be of the `name:type` form, so there's no confusion. But still, maybe `plot` looks too plain? Let me know if you have a better idea. I didn't want a whole new box, because they are quite similar otherwise.

app/com/lynxanalytics/biggraph/frontend_operations/WorkflowOperations.scala
```diff
@@ -609,8 +609,14 @@ class WorkflowOperations(env: SparkFreeEnvironment) extends ProjectOperations(en
...
+              val html = PythonUtilities.deriveHTML(params("code"), pythonInputs, project)
+              // The output is called "graph" to preserve compatibility.
```
darabos, line 614: Any anyway, a "graph" can mean a technical illustration, right?

app/com/lynxanalytics/biggraph/graph_operations/DeriveHTMLPython.scala
```diff
@@ -0,0 +1,49 @@
+// Creates an HTML output using Python executed on Sphynx.
+// For security, this output should not be freely controlled by the user.
+// Instead it's used for returning plots encoded in HTML.
```
darabos, line 3: You can see here how torn I am between making this something extremely generic and limiting it to plots. We could call it `DerivePlotPython`. But what if we later add another Python operation that returns HTML? We would have to copy-paste the whole thing. Or I guess we could subclass it. That doesn't sound so bad actually. What do you think?

sphynx/python/derive_html.py
```diff
@@ -0,0 +1,51 @@
+'''Run user code.'''
```
darabos, line 1: Sorry, forgot to update this. The first half of this is just a copy of `derive.py`.

sphynx/python/derive_html.py
```diff
@@ -0,0 +1,51 @@
...
+  import traceback
+  import sys
+  a, b, c = sys.exc_info()
+  traceback.print_exception(a, b, c.tb_next)
+  sys.exit(1)
+
+f = io.BytesIO()
+plt.savefig(f, format='svg')
+data = base64.b64encode(f.getvalue()).decode('ascii')
+op.output_scalar('sc', f'<img src="data:image/svg+xml;base64, {data}">')
```
darabos, line 51: So the plot is encoded in the HTML.

Because we do this after running the user code, this is not entirely secure. The user could write out their own HTML fragment with a line just like this, and then change `op.output_scalar` to do nothing.

At this point I should clarify what kind of security risk I'm talking about.

If the user can inject arbitrary HTML into the workspace, they can then ask another user to view their workspace. If their injected HTML contains JavaScript, that will be executed under the user credentials of the user viewing the page. ([Cross-site scripting](https://en.wikipedia.org/wiki/Cross-site_scripting))

What we could do to avoid the security issue described in the second paragraph, is to let the Python code just write out the image. Then the Go code would wrap it in an `` tag like on this line. The user would have no way to trick the Go code.

darabos (commented):
I still have to add tests and docs. But the functional part is ready for review!

sphynx/python/derive_html.py
darabos: > At this point I should clarify what kind of security risk I'm talking about.

The most common method to avoid XSS is to serve content from a different domain. (e.g. githubusercontent.com) We can't exactly do that, but we can still use an `<iframe>`. It turns out iframe has a `sandbox` attribute that makes it behave as if its content is from a different domain. So cool! It also disables scripts and a lot of things by default, which is fine for us for now. But this opens the door to secure interactive Python plots later on.

darabos:
Some interactive visualization libraries now work too. This screenshot shows Deck.gl, Plotly, and py3Dmol:
image

The three modes supported now are:
- `matplotlib`: You use it as normal. Don't call `plt.show()`. We take the plot from the global state.
- `ipywidget`: We also take it from the global state. These plots generally involve Python code in the interactive part, which doesn't work without a Python backend. I almost got NGLView working, but not entirely. Simpler examples work.
- `html`: The most flexible. All three in the screenshot use this. You just put some HTML in a variable called `html` and that's it.

web/app/scripts/workspace/html-state.html
```diff
@@ -0,0 +1,6 @@
+<uib-progressbar ng-if="!content.$resolved" value="100" class="progress-striped active">
+Loading...
+</uib-progressbar>
+<div class="col-sm-12" ng-show="content.$resolved">
+  <div trusted-html="content.string"></div>
+</div>
```
lacca0: Add a newline please.

lacca0 (approved):

web/app/scripts/workspace/html-state.html
darabos: > Add a newline please.

Weird. The missing newline was fixed 20 days ago. Did you have this browser tab open that long? 😄  Here's the commit history of this file: https://github.com/lynxkite/lynxkite/commits/darabos-html/web/app/scripts/workspace/html-state.html

Anyway, the newline is good now. 👍 Thanks for the review!

app/com/lynxanalytics/biggraph/graph_operations/DeriveHTMLPython.scala
darabos: > I would make a subclass if I were you.

👍 If we need another Python+HTML operation later, we'll do that. (I think creating a big automatic graph analysis report is one idea that could use this. We'll see.)

darabos:
> I still have to add tests and docs. But the functional part is ready for review!

Oops! In my excitement over the approvals I forgot about the tests and docs. Sorry. Will add them tomorrow.

PR #299, opened 2022-11-03
Delete pid files

### LINKED ISSUE #251: Remove PID file
LynxKite checks a "PID" file on startup and won't start if it exists. I guess we were worried about running duplicate processes at some point, or we just saw a cool program doing this and wanted to imitate it. But in practice it's pointless. We will fail to start up anyway if the port is taken. And I often find myself in the last week having to delete the PID file from Docker containers.

That's a bit of a surprise, since it's supposed to be automatically deleted by https://github.com/lynxkite/lynxkite/blob/5.0.0/docker/run.sh#L4. I guess something changed. (Probably the CWD.)

lacca0:
Resolves #251


darabos:
You have conflicts.

darabos (commented):
Thanks, that's a good start! But LynxKite still creates a pidfile, right? It's called `RUNNING_PID` in the current directory. Looks like it's coming from [Play](https://www.playframework.com/documentation/2.8.x/ProductionConfiguration).

To make sure we don't miss anything, you could try running LynxKite twice, on two different ports.

run.sh
```diff
@@ -5,5 +5,4 @@ cd $(dirname $0)
 tools/install_spark.sh
 make backend
 . ~/.kiterc
-$SPARK_HOME/bin/spark-submit \
-  target/scala-2.12/lynxkite-0.1-SNAPSHOT.jar
+spark-submit --conf 'spark.driver.extraJavaOptions=-Dplay.http.secret.key=$KITE_APPLICATION_SECRET' target/scala-2.12/lynxkite-0.1-SNAPSHOT.jar
```
darabos, line 8: Did you add this flag by accident?
```suggestion
spark-submit target/scala-2.12/lynxkite-0.1-SNAPSHOT.jar
```

darabos (approved):
That's all? No more pid files? Have you tried running two instances?

Thanks!

PR #302, opened 2022-11-17
Tests and docs for Python plotting

darabos:
Leftover work from #297. I've also added support for plotting with table inputs. It was easier to do than explaining in the user guide why plotting is only available for graph inputs. 😅


darabos:
I spent half a day figuring out `saveWidget()`. 😭 But now plotting in R is even better than plotting in Python. Maybe there's a universal object-to-HTML function in Python too? Oh well, we can add it later if we find it.

lacca0 (approved):

Issue #238, opened 2022-05-24
Rewrite the frontend tests with Playwright


darabos:
The Protractor tests fail a lot of the time. Curiously it's often just a few test cases and they fail with:

```
unhandled inspector error: {"code":-32000,"message":"Inspected target navigated or closed"}
```
We're not the first to experience this. (https://github.com/angular/protractor/issues/5347)

Ideas:
- The only remedy described is adding retries. That's not so bad, because it probably only happens on `browser.get()` which we only call in the `empty splash page` test case. But I'm not sure we can add retries without rewriting the tests to use await/async. (Which would be a good thing anyway, but it's a huge effort.)
- Try upgrading things. I have no idea what's our Chromedriver version.

darabos:
We just noticed the frontend code has some references to `environment.protractor`. Let's not forget to remove those!

darabos:
Only `center-picker-test.js` left now! It's an important test. But it's complex, commented out, and we plan to redesign the center picker anyway as part of the visualization revamp. So let's leave it dead for now.

PR #307, opened 2022-11-21
Start converting tests to Playwright

darabos:
First step for #238. Includes a conversion of `web/test/tests/splash-page-test.js`.

I haven't updated the docs yet. You can see the CI setup in the Earthfile. For local development you install Playwright the same way. Then you run LynxKite with `run.sh` and run the tests with `yarn playwright test`. There's a nice debug utility you can start with `yarn playwright debug`. If you hit an issue that you can't reproduce with `debug`, you can run the tests with `yarn playwright test --trace on`. This will save a trace that you can replay to see what happened.

I want to set up the trace on GitHub Actions too. The trace will have to be saved as an Earthly local artifact, then saved as a GitHub Actions artifact. This will be fantastic to track down any flaky tests. Or maybe we just won't have flaky tests at all!


web/tests/splash.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (await page.splash.directory('automated-tests').isVisible()) {
+    await page.splash.deleteDirectory('automated-tests');
+  }
+  await page.splash.newDirectory('automated-tests');
+  await page.splash.expectNumWorkspaces(0);
+  await page.splash.expectNumDirectories(0);
+  return page;
+}
+
+test.describe.configure({ mode: 'serial' });
```
darabos, line 28: The Protractor tests have an intricate test framework implemented in [`web/test/declarative.js`](https://github.com/lynxkite/lynxkite/blob/main/web/test/declarative.js). Instead of a simple independent/serial execution this allows a tree of tests, where some tests preserve the state and some tests change it.

This is made possible by the `discardAllReallyIMeanIt` endpoint, which resets LynxKite completely and provides a safe starting state. I'm not fond of this, because I often have some nice workspaces for testing stuff I'm working on, and then the tests blow it away.

My hope is that we can ditch all this. We can get to a known state by (re)creating a directory for the tests. We can write each test file to be runnable on its own. They can still share code: another test could `newSplash()` from this file and just call it. We don't need to build a framework for this.

If we put the test stuff for each test file into a separate LynxKite directory, perhaps we could one day run those in parallel. But it's fine not to worry about this for now.

darabos:
On my laptop when I run all the tests, something goes wrong when "SQL runs nice on belongs to reached from project and segmentation" tries to add the "Create vertices" box. It clicks on it but the popup doesn't come up. I think this may be an actual LynxKite bug.

A difference between Protractor and Playwright is that Protractor waited for network requests to finish before taking the next action. Playwright doesn't care. So it may be clicking the box before the `getOrSetWorkspace` call has finished. Which is fine! A user wouldn't wait either! And I have actually experienced clicking boxes and not getting popups. Fixing things for Playwright also fixes things for users.

darabos:
It passed on GitHub Actions. I guess let's merge it! (After review.) Then we can start converting more tests and fixing things as we go.

web/tests/sql.spec.ts
```diff
@@ -0,0 +1,467 @@
...
+  await tableBrowser.expectNode([4], 'two.edge_attributes', '`two.edge_attributes`');
+  await tableBrowser.expectNode([5], 'two.edges', '`two.edges`');
+  await tableBrowser.expectNode([6], 'two.graph_attributes', '`two.graph_attributes`');
+  await tableBrowser.expectNode([7], 'two.vertices', '`two.vertices`');
+  await tableBrowser.toggleNode([3]);
+  await tableBrowser.expectNode([3, 0], '*ALL*');
+  await tableBrowser.expectNode([3, 1], 'age (Double)', '`age`');
+});
+
+/*
```
darabos, line 238: These are commented-out in the original. But let's fix what we can and delete what we no longer need during this migration.

darabos: > These are commented-out in the original. But let's fix what we can and delete what we no longer need during this migration.

I've reviewed them. Nothing worth fixing I think. I've deleted them all.

web/tests/lynxkite.ts
```diff
@@ -0,0 +1,1615 @@
...
+    const after = boxData.after;
+    const inputs = boxData.inputs;
+    const params = boxData.params;
+    await this.page.evaluate(boxData => {
+      $(document.querySelector('#workspace-drawing-board')).scope().workspace.addBox(
+        boxData.name,
+        { logicalX: boxData.x, logicalY: boxData.y },
+        { boxId: boxData.id });
+    }, boxData);
+    // Wait for the backend to save this box.
```
darabos, line 198: Alternatively we could make it so that it's okay to click a box before it's saved. I've filed https://github.com/lynxkite/lynxkite/issues/308.

darabos (commented):
I've managed to track down some reliability issues. Looks rock solid for me now!

Earthfile
```diff
@@ -1,4 +1,4 @@
-VERSION --use-copy-include-patterns 0.6
+VERSION --use-copy-include-patterns --try 0.6
```
lacca0, line 1: Cannot find this option at docs.earthly.dev. What does it do?

Earthfile
```diff
@@ -25,24 +25,24 @@ sbt-deps:
 
 npm-deps:
   COPY web/package.json web/yarn.lock web/
-  RUN cd web; yarn --frozen-lockfile
+  RUN cd web && yarn --frozen-lockfile
```
lacca0, line 28: Good idea!

tools/with_lk.sh
```diff
@@ -18,46 +21,37 @@ cat > "$KITE_USERS_FILE" <<EOF
   "isAdmin" : false
 } ]
 EOF
-KITE_DEPLOYMENT_CONFIG_DIR="$(dirname $0)/../test"
-KITE_DEPLOYMENT_CONFIG_DIR=$(realpath "${KITE_DEPLOYMENT_CONFIG_DIR}")
-export HTTP_PORT=$[ 9100 + RANDOM % 100 ]
-export HTTPS_PORT=$[ 9200 + RANDOM % 100 ]
+export KITE_HTTP_PORT=$[ 9100 + RANDOM % 100 ]
```
lacca0, line 24: Why did we need to make these ports random at the first place?

.github/workflows/earthly-test.yml
```diff
@@ -33,13 +34,14 @@ jobs:
         fi
         git checkout -b "$branch" || true
     - name: Download latest earthly
-      run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/download/v0.6.26/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
+      run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/download/v0.6.30/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
```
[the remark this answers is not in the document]
darabos: > Why can't we use /earthly/releases/latest/...?

With a fixed version the behavior is more predictable. With `/latest` there would always be a chance that something broke because of a new Earthly release.

(I'm in favor of a lockfile for the Conda dependencies too, for the same reason.)

But maybe we can switch to `/latest` when Earthly becomes stable enough, e.g. 1.0.

Earthfile
darabos: > Cannot find this option at docs.earthly.dev. What does it do?

Yeah, it's pretty new. See https://github.com/earthly/earthly/issues/587#issuecomment-1255633387.

Unfortunately it only works for files under 64 kB at the moment. (https://github.com/earthly/earthly/issues/2452) We will only start getting test reports when a new Earthly is released with the fix.

Earthfile
darabos: > Good idea!

For `cd` it doesn't make much difference. But I used `;` instead of `&&` at some other step and ended up ignoring a failure as a result. I took revenge on all semicolons.

tools/with_lk.sh
darabos: > Why did we need to make these ports random at the first place?

I think it's to make it more convenient for local testing. You usually run a LynxKite on port 2200 that you're working with. You can use that for testing. This script is for starting a different, brand new LynxKite just for the test.

For testing in Earthly it doesn't matter. A fixed port would be fine.

lacca0 (approved):
Thanks!

darabos:
I was considering whether to merge this before the release. I think it's fine. The `enqueue` fix is a small change. Playwright likes it. It looks okay in manual testing. I've run the Protractor tests and they pass too.

PR #309, opened 2022-11-30
Fix state-toolbar ID for tests

darabos:
Doesn't look like an element can have multiple IDs like this. I don't know how this ever worked. For me this fixed the Protractor tests.


darabos:
@lacca0 says she approves.

PR #317, opened 2022-12-13
Start converting choice-params test

darabos:
In the end I didn't add a custom matcher. 😞 It sounded cool, but 1) it's not useful this test, since we're comparing two histograms here, and 2) I can't really make it better than taking the histogram in the current format and then comparing it. We _could_ do the comparison element-by-element, but seeing "5 does not equal 6" is actually worse than seeing the two complete histograms that are mismatched. At the `toEqual()` matcher may already be doing better than that.


lacca0 (approved):

darabos:
This file is unfinished, but let's merge what we have at the moment.

PR #318, opened 2022-12-15
Finish converting the choice-params tests

darabos:
Where #317 left off.

### DIFF
--- a/web/tests/choice-params-with-default.spec.ts
+++ b/web/tests/choice-params-with-default.spec.ts
@@ -8,7 +8,7 @@ test.beforeAll(async ({ browser }) => {
   await workspace.addBox({ id: 'ex0', name: 'Create example graph', x: 100, y: 100 });
 });
 
-test('test pagerank default choice values', async () => {
+test('pagerank default choice values', async () => {
   await workspace.addBox({
     id: 'pr1',
     name: 'Compute PageRank',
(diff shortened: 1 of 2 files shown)


web/tests/choice-params-with-default.spec.ts
```diff
@@ -8,7 +8,7 @@ test.beforeAll(async ({ browser }) => {
   await workspace.addBox({ id: 'ex0', name: 'Create example graph', x: 100, y: 100 });
 });
 
-test('test pagerank default choice values', async () => {
+test('pagerank default choice values', async () => {
```
darabos, line 11: The original strings are state names, and even as state names they are not very clear. I'm thinking how we could improve them. Here I'm trying to not change much, but at least make the name specify what we are testing. We're testing these default choice values. Then we are testing editing them. It makes a bit more sense in my mind.

web/tests/choice-params-with-default.spec.ts
```diff
@@ -31,3 +31,33 @@ test('test pagerank default choice values', async () => {
...
+  await workspace.addBox({ id: 'ex', name: 'Create example graph', x: 100, y: 100 });
+  await workspace.addBox({
+    id: 'discard', name: 'Discard vertex attributes', x: 100, y: 200, after: 'ex'
+  });
+  const box = await workspace.openBoxEditor('discard')
+  await box.moveTo(500, 100);
+  const state = await workspace.openStateView('discard', 'graph');
+  await state.moveTo(500, 400);
+  const attrs = state.left.side.locator('entity[kind="vertex-attribute"]');
+  await expect(attrs).toHaveText(['age', 'gender', 'id', 'income', 'location', 'name'], { useInnerText: true });
```
darabos, line 56: `useInnertText` comes from https://github.com/microsoft/playwright/issues/19072. I tried to make it the default by overriding `toHaveText` with `expect.extend`. Then I tried adding a new matcher. But it's not easy to extend existing matchers like this. (See https://github.com/facebook/jest/issues/6243 and https://github.com/facebook/jest/issues/10329.)

I went with just typing out `useInnerText` for now. We can look into cleaning this up later.

web/tests/lynxkite.ts
```diff
@@ -425,14 +416,8 @@ export class BoxEditor extends PopupBase {
...
-    const param = this.element.locator('div#param-' + paramName + ' input');
-    expect(param.getAttribute('value')).toBe(expectedValue);
-  }
-
-  expectSelectParameter(paramName, expectedValue) {
-    const param = this.element.locator('div#param-' + paramName + ' select');
-    expect(param.getAttribute('value')).toBe(expectedValue);
+  getParameter(paramName, tag = 'input') {
+    return this.element.locator(`div#param-${paramName} ${tag}`);
   }
```
darabos, line 421: The old code had lots of functions that make assertions. I think if the assertion is simple enough, then it's easier to understand if it's made in the test. (Not in `lynxkite.ts`.)

lacca0 (approved):

PR #321, opened 2022-12-15
Convert plot-test.js and metaeditor-test.js

darabos:


lacca0 (approved):

darabos:
Okay, I finally have a clue about why the test is failing.

```
Syntax error at or near 'vertices': extra input 'vertices'(line 1, pos 76)

== SQL ==
select edge_comment, src_name from edges order by edge_commentselect * from vertices
----------------------------------------------------------------------------^^^
```

Looks like no matter how much I hoped a simple `fill()` will do the trick, the code editor box (ACE) needs a bit more attention.

PR #323, opened 2023-01-10
Add delete box test 

lacca0:


web/tests/lynxkite.ts
```diff
@@ -427,7 +427,7 @@ export class BoxEditor extends PopupBase {
   }
 
   isPresent() {
```
darabos, line 429: I think it's better to delete this. You're not even using it.

Please delete `boxExists`/`boxPopupExists` too. Instead use `getBox` and `getBoxEditor` with a `toBeVisible` check. Thanks!

web/tests/delete-test.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    y: 300,
+    after: 'pr1',
+    params: { name: 'page_rank_incoming', direction: 'incoming edges', iterations: '1' },
+  });
+});
+
+test('delete boxes', async () => {
+  await workspace.clickBox('pr2');// To check that only popups for deleted boxes are closed.
+  await workspace.deleteBoxes(['pr1', 'ex0']);
+  expect(await workspace.boxExists('ex0')).toEqual(false);
```
darabos, line 30: Instead of `expect(await x.isVisible()).toEqual(true)` please write `await expect(x).toBeVisible()`.

web/tests/delete-test.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    name: 'Compute PageRank',
+    x: 100,
+    y: 300,
+    after: 'pr1',
+    params: { name: 'page_rank_incoming', direction: 'incoming edges', iterations: '1' },
+  });
+});
+
+test('delete boxes', async () => {
+  await workspace.clickBox('pr2');// To check that only popups for deleted boxes are closed.
```
darabos, line 28: Use `openBoxEditor()`. It has a clearer intention than `clickBox()`. It also returns the popup, so you can later use that in the assertion.

Also open the popup for `pr1`. Like the comment says, we want to see that the popups for deleted boxes are gone. But in your test the popup for the deleted box is not even open.

web/tests/delete-test.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+  await workspace.deleteBoxes(['pr1', 'ex0']);
+  await expect(workspace.board.locator('.box#ex0')).not.toBeVisible();
```
darabos, line 30: Why not use `getBox()` and `getBoxEditor()` like I suggested?

```suggestion
  await expect(workspace.getBox('ex0')).not.toBeVisible();
```

darabos (commented):
Almost there! Thanks!

darabos (approved):
You have conflicts again! You're doing something funny with Git, aren't you?

PR #325, opened 2023-01-12
Rewrite drag selection test

lacca0:
Do I need to rewrite selectArea()?

### DIFF
--- a/tools/with_lk.sh
+++ b/tools/with_lk.sh
@@ -39,7 +39,7 @@ export KITE_ALLOW_NON_PREFIXED_PATHS=true
 cd $REPO
 spark-submit \
   --conf "spark.driver.extraJavaOptions=-Dplay.http.secret.key=SECRET-TEST-TEST-TEST-TEST -Dhttps.keyStore=$KITE_HTTPS_KEYSTORE -Dhttps.keyStorePassword=$KITE_HTTPS_KEYSTORE_PWD" \
-  target/scala-2.12/lynxkite-0.1-SNAPSHOT.jar &
+  target/scala-2.12/lynxkite-0.1-SNAPSHOT.jar > /dev/null 2>&1 &
 KITE_PID=$!
 cd -
 function kill_backend {
--- a/web/playwright.config.ts
+++ b/web/playwright.config.ts
@@ -13,7 +13,7 @@ const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
   retries: process.env.CI ? 2 : 0,
   workers: 1,
-  reporter: process.env.CI ? [['github'], ['html']] : 'list',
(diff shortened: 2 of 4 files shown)


darabos (approved):
Awesome! 👍 

> Do I need to rewrite selectArea()?

Just deleting it is fine! Thanks!

Earthfile
```diff
@@ -145,7 +145,7 @@ frontend-test:
   COPY conf/kiterc_template conf/
   COPY test/localhost.self-signed.cert* test/
   ENV CI true
-  RUN cd web && ../tools/with_lk.sh yarn playwright test || touch failed
+  RUN cd web && ../tools/with_lk.sh > /dev/null 2>&1 yarn playwright test || touch failed
```
darabos, line 148: Come on, I even said to add the redirect _in_ `with_lk.sh`. I also said to test it locally. But even if you didn't, you can look at the GitHub Actions output and see this is wrong.

I think you're very confused about something, so let me break down what's happening:

```python
RUN
  cd web
&&
  ../tools/with_lk.sh yarn playwright test
||
  touch failed
```

Three commands with logical operators between them. Each command is a space-separated list where the first item is an executable (or shell built-in) and the rest are its arguments. Specifically in the second command we run `with_lk.sh` with the `yarn`, `playwright`, `test` arguments.

What happens if you stick `> /dev/null 2>&1` in the middle of the argument list? I've never seen anything like this, but it looks like it's the same as sticking it at the end:

```bash
$ echo x > y z
$ cat y
x z
```

Wild!

Perhaps you don't know what `with_lk.sh` does? It has a comment explaining it and it's not a lot of code. But maybe we never discussed it. It starts LynxKite in the background, runs the command it got as an argument, then shuts down LynxKite.

What your change does is mute this whole step. So LynxKite is successfully muted, but you're also muting Playwright!

I'm certainly dumber and slower when I have a cold, so all dumbness and slowness for both of us can be excused this week. You use Linux as your primary system, we built the Earthly setup and the Playwright setup together, and you reviewed major changes in `with_lk.sh`, so I think you understand all this. But let me know if we need to go over some things.

darabos:
×±···°·······×±·················×±···

The test passed, but it had to retry 3 tests. If you download the test results artifact from https://github.com/lynxkite/lynxkite/actions/runs/3966721454 you can see which ones. Two of them were waiting for histograms to load, but it took too long. We may need to increase the timeouts for that or make LynxKite faster. In the third case the SQL box output didn't open after clicking. That may be a UI issue that we will have to fix at some point.

But for now the retries are okay. Let's just merge this already.

Also the GitHub Actions artifact is the wrong stuff. It's just the traces. We need the `playwright-report` directory, not the `test-results` directory. They have basically the same stuff, but the other one has the nice HTML overview with links.

PR #327, opened 2023-01-12
Use DataFrames from PySpark without writing them to disk

darabos:
I'll add a test! For now I've tested it manually like this:

```
from pyspark.sql import Row
import lynx.kite
lk = lynx.kite.LynxKite(spark=spark)
df = spark.createDataFrame([Row(name='Adam', age=23), Row(name='Eve', age=34), Row(name='Bob', age=45)])
print(lk.from_spark(df).sql('select * from input where age < 40').df())
```

It's so cool! I'm also going to add the other direction in another PR: LynxKite table to PySpark DataFrame.

### DIFF
--- a/python/remote_api/src/lynx/kite.py
+++ b/python/remote_api/src/lynx/kite.py
@@ -340,6 +340,8 @@ class LynxKite:
     spark = SparkSession.builder.config('spark.jars', 'lynxkite-X.X.X.jar').getOrCreate()
   '''
 
+  _managed = None
+
   def __init__(self, username: Optional[str] = None, password: Optional[str] = None, address: Optional[str] = None,
                certfile: Optional[str] = None, oauth_token: Optional[str] = None, signed_token: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


python/remote_api/src/lynx/kite.py
```diff
@@ -865,11 +865,30 @@ def from_df(self, df, index=True):
...
+    Converts a Spark DataFrame to a LynxKite table.
+    Costs almost nothing if LynxKite was started in the same SparkSession.
+    '''
+    if self._lynxkite:
+      # The DataFrame is in the same JVM! We can use it directly.
+      guid = self._lynxkite.importDataFrame(df)
+      print('guid:', guid)
+      return SingleOutputAtomicBox(
+          self.box_catalog(), self, 'importParquet', inputs={}, parameters={
+              'filename': 'from PySpark',
```
darabos, line 878: I strategically chose this filename so that the box looks like this:

image

lacca0 (approved):
Elegant! No questions:)

PR #329, opened 2023-01-23
Fix built-ins loading

### LINKED ISSUE #324: Built-in custom boxes are not loaded on startup
We missed this because the long-running instances have already loaded them in earlier versions. But in a clean install, they are not there. This is probably fallout from all the startup changes.

By the way, we could them inside the `custom_boxes` directory instead of a separate `built-ins`. One less special top-level directory.

darabos:
Fixes #324.

I spent a crazy amount of time trying to add `built-ins` as a resource. The tricky part is that if I just add it, the files will be in the root of the jar. I wanted them inside `built-ins`. The `mappings` setting is supposed to be used for that, but no matter how much I upgraded SBT it didn't work. With `stage` the files got in the right place, but not with `assembly`.

In the end I put `built-ins` inside a `resources` directory. This is simple and easy. Why didn't I think of it sooner!


Earthfile
```diff
@@ -64,7 +64,7 @@ app-build:
   COPY +grpc/proto app/com/lynxanalytics/biggraph/graph_api/proto
   COPY conf conf
   COPY app app
-  COPY built-ins built-ins
+  COPY resources resources
```
lacca0, line 67: Could you remind me why did you need to rename the directory?

lacca0 (approved):

Earthfile
darabos: > Could you remind me why did you need to rename the directory?

It's in the PR description!

lacca0: Ah, so now we have a "built-ins" directory inside a "resources" directory?

lacca0 (approved):

PR #331, opened 2023-01-23
Add help page test

lacca0:
Fix the test according to these comments:
https://github.com/lynxkite/lynxkite/commit/9ee24794ffca2f24385918c5a0dd73207f2bf853#r97679988


web/tests/fork-workspace.spec.ts
```diff
@@ -8,11 +8,10 @@ let workspace: Workspace;
...
-    await workspace.page.pause();
     await workspace.page.locator('#save-workspace-as-starter-button').click();
-    await workspace.page.locator('#save-workspace-as-input').locator('input').fill('test-example-fork');
-    await workspace.page.locator('#save-workspace-as-input').locator('#ok').click();
-    await workspace.page.waitForURL('**/#/workspace/test-example-fork');
+    await workspace.page.locator('#save-workspace-as-input input').fill('automated-tests/subdirectory/test-example-fork');
+    await workspace.page.locator('#save-workspace-as-input #ok').click();
+    await workspace.page.waitForURL('**/#/workspace/automated-tests/subdirectory/test-example-fork');
     await workspace.close();
     await splash.expectWorkspaceListed('test-example-fork');
```
darabos, line 16: Please assert that we're in `automated-tests/subdirectory`.

darabos (approved):
Thanks! It's not a refactor though. Maybe change the title?

web/tests/help-page.spec.ts
```diff
@@ -0,0 +1,62 @@
+//Tests the help page
+import { test, expect } from '@playwright/test';
+import { TableBrowser, Workspace } from './lynxkite';
+
+let workspace: Workspace;
+
+async function checkExactlyOneCurrent() {
+    // There is only one highlighted item:
+    await expect(await workspace.page.locator('span.find-highlight-current').count()).toBe(1);
```
darabos, line 9: `toHaveCount`

web/tests/help-page.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+async function checkCurrentHighlight(expectedPos) {
+    await checkExactlyOneCurrent();
+    // Check correct item:
+    const cssQuery = 'span.find-highlight.find-highlight-current.find-highlight-' + expectedPos;
+    await expect(workspace.page.locator(cssQuery)).toBeVisible();
+}
+
+test.beforeAll(async ({ browser }) => {
+    workspace = await Workspace.empty(browser);
+});
```
darabos, line 21: You don't need a workspace here, right? Delete this.

web/tests/help-page.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+test('search box in help', async () => {
+    await workspace.page.goto('#/help');
+    const field = await workspace.page.locator('#find-in-page-text');
+    await expect(field).toBeVisible;
+    // Search for the phrase "user".
+    await field.fill('user');
+    // Expect more than 5 matches.
+    await workspace.page.locator('span.find-highlight').first().waitFor();
```
darabos, line 30:
```suggestion
    await expect(workspace.page.locator('span.find-highlight').first()).toBeVisible();
```

This waits too, but also adds an assertion and it's more like all the other checks.

darabos (approved):
Looking great!

web/tests/help-page.spec.ts
lacca0: I think I need it to access the Page interface later. Do you suggest using the Splash class for it?
await workspace.page.goto('#/help'); 
vs 
await splash.page.goto('#/help');

darabos: > I think I need it to access the Page interface later.

Of course. Just use the page you get from Playwright. Please look at the docs at some point. https://playwright.dev/docs/writing-tests#first-test

web/tests/help-page.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    // Text not found:
+    await field.click();
+    await field.fill('qwertyui');
+    await expect(await workspace.page.locator('span.find-highlight').count()).toBe(0);
+});
+
+test('scroll position in help', async () => {
+    await workspace.page.goto('#/help#graph-visualization');
+    const help = await workspace.page.locator('#whole-help');
+    await expect(help).not.toHaveAttribute('scrollTop', '0');
```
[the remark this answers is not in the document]
lacca0: Well I get 'null' in both places.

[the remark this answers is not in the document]
darabos: > Well I get 'null' in both places.

Good thing we checked then! I think it probably happens because `scrollTop` is not an HTML attribute. I'm surprised `getAttribute` worked in the original code.

Maybe you find a better solution. But if all else fails, you can always use `help.evaluate(...)` to get the scroll position as you would in the browser. (Maybe `help.evaluate(e => e.scrollTop)`?)

By the way, you don't need `await` on the line above.

web/tests/fork-workspace.spec.ts
```diff
@@ -8,11 +8,11 @@ let workspace: Workspace;
 test('fork workspace', async function ({ browser }) {
     splash = await Splash.open(browser);
     workspace = await splash.openNewWorkspace('test-example');
-    await workspace.page.pause();
     await workspace.page.locator('#save-workspace-as-starter-button').click();
-    await workspace.page.locator('#save-workspace-as-input').locator('input').fill('test-example-fork');
-    await workspace.page.locator('#save-workspace-as-input').locator('#ok').click();
-    await workspace.page.waitForURL('**/#/workspace/test-example-fork');
+    await workspace.page.locator('#save-workspace-as-input input').fill('automated-tests/subdirectory/test-example-fork');
```
darabos, line 12: Instead of including `automated-tests` here, could you please use `ROOT` from `lynxkite.ts`? There's no big practical reason for it, but that's what we did in all the other files. Thanks.

web/tests/fork-workspace.spec.ts
```diff
@@ -8,11 +8,11 @@ let workspace: Workspace;
 test('fork workspace', async function ({ browser }) {
     splash = await Splash.open(browser);
     workspace = await splash.openNewWorkspace('test-example');
```
darabos, line 10: Please change the workspace name to be specific to this test. (As long as we have to enter a name, why not use distinct names? It makes troubleshoot a bit less confusing I think.)

web/tests/fork-workspace.spec.ts
```diff
@@ -8,11 +8,11 @@ let workspace: Workspace;
...
-    await workspace.page.pause();
     await workspace.page.locator('#save-workspace-as-starter-button').click();
-    await workspace.page.locator('#save-workspace-as-input').locator('input').fill('test-example-fork');
-    await workspace.page.locator('#save-workspace-as-input').locator('#ok').click();
-    await workspace.page.waitForURL('**/#/workspace/test-example-fork');
+    await workspace.page.locator('#save-workspace-as-input input').fill('automated-tests/subdirectory/test-example-fork');
+    await workspace.page.locator('#save-workspace-as-input #ok').click();
+    await workspace.page.waitForURL('**/#/workspace/automated-tests/subdirectory/test-example-fork');
     await workspace.close();
+    await workspace.page.waitForURL('**/#/dir/automated-tests/subdirectory');
```
darabos, line 16: 👍

web/tests/help-page.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    await page.goto('#/help');
+    const field = page.locator('#find-in-page-text');
+    const help = page.locator('#whole-help');
+    expect(await help.evaluate(e => e.scrollTop)).toBe(0);
+    await expect(field).toBeVisible();
+    // Search for the phrase "user".
+    await field.fill('user');
+    // Expect more than 5 matches.
+    await expect(page.locator('span.find-highlight').first()).toBeVisible();
+    expect(await page.locator('span.find-highlight').count()).toBeGreaterThan(5);
```
darabos, line 26: Could we always use `await expect(...)` instead of `expect(await ...)`? Or does that not work everywhere?

darabos (approved):
Awesome, thanks!

web/tests/fork-workspace.spec.ts
lacca0: Didn't you delete this test in main? #238 says it's merged into splash.spec.ts.

darabos: Looks like I'm deleting it in #341. Which I've sent out more than a week ago! Perhaps it's time to start reviewing pull requests? 😈 (I've already merged a few where I gave up waiting. Please review these too: #336, #337, #338, #339, #345. Thanks!)

web/tests/help-page.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    // Check correct item:
+    const cssQuery = 'span.find-highlight.find-highlight-current.find-highlight-' + expectedPos;
+    await expect(page.locator(cssQuery)).toBeVisible();
+}
+
+test('search box in help', async ({ page }) => {
+    await page.goto('#/help');
+    const field = page.locator('#find-in-page-text');
+    const help = page.locator('#whole-help');
+    expect(help.evaluate(e => e.scrollTop)).toBe(0);
```
darabos, line 20: This needs an `await`. `evaluate` returns a promise. I think you can do `await expect(promise).toBe(whatever)`. But if that doesn't work, then you have to do `expect(await promise).toBe(whatever)`.

If you have the patience to look into our ESLint setup, this is the rule that would help catch these mistakes: [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises/)

darabos (changes requested):

web/tests/help-page.spec.ts
darabos: > Could we always use `await expect(...)` instead of `expect(await ...)`? Or does that not work everywhere?

Check out what I just found! This is the recommendation in the Playwright docs too! It's not just me! https://playwright.dev/docs/best-practices#use-web-first-assertions

darabos: > Check out what I just found! This is the recommendation in the Playwright docs too! It's not just me! https://playwright.dev/docs/best-practices#use-web-first-assertions

On second thought, it's not just about where you put the `await`, but about using the smart built-in assertions. Using `toHaveCount(x)` instead of `.count()).toBe(x)`.

We could actually do that. Instead of checking that the count is >5, check that the 6th item is visible!

```suggestion
    await expect(page.locator('span.find-highlight').first()).toBeVisible();
    await expect(page.locator('span.find-highlight').nth(6)).toBeVisible(); // We have at least 6 matches.
```

web/tests/help-page.spec.ts
lacca0: > Looks like I'm deleting it in #341. 
I think you said to delete those on merge.

web/tests/help-page.spec.ts
darabos: This would still be good. :point_up:

darabos (approved):

web/tests/fork-workspace.spec.ts
darabos: > Looks like I'm deleting it in #341. Which I've sent out more than a week ago! Perhaps it's time to start reviewing pull requests? 😈 (I've already merged a few where I gave up waiting. Please review these too: #336, #337, #338, #339, #345. Thanks!)

This would also still be good. :point_up: 🙂

PR #333, opened 2023-01-23
Convert upload test

darabos:
For #238.

### DIFF
--- a/web/tests/choice-params-with-default.spec.ts
+++ b/web/tests/choice-params-with-default.spec.ts
@@ -4,7 +4,7 @@ import { Workspace } from './lynxkite';
 
 let workspace: Workspace;
 test.beforeAll(async ({ browser }) => {
-  workspace = await Workspace.empty(browser);
+  workspace = await Workspace.empty(await browser.newPage());
   await workspace.addBox({ id: 'ex0', name: 'Create example graph', x: 100, y: 100 });
 });
 
--- a/web/tests/data/upload_test.csv
+++ b/web/tests/data/upload_test.csv
@@ -1,0 +1,5 @@
+"name"
+"Adam"
+"Eve"
+"Bob"
+"Isolated Joe"
--- a/web/tests/drag-select.spec.ts
+++ b/web/tests/drag-select.spec.ts
@@ -4,26 +4,26 @@ import { Workspace } from './lynxkite';
 
 let workspace: Workspace;
 test.beforeAll(async ({ browser }) => {
-    workspace = await Workspace.empty(browser);
-    await workspace.addBox({ id: 'ex0', name: 'Create example graph', x: 100, y: 100 });
-    await workspace.addBox({
-      id: 'pr1',
-      name: 'Compute PageRank',
-      x: 100,
-      y: 200,
-      after: 'ex0',
-    });
-    await workspace.addBox({
-      id: 'pr2',
-      name: 'Compute PageRank',
-      x: 100,
-      y: 300,
-      after: 'pr1',
-    });
+  workspace = await Workspace.empty(await browser.newPage());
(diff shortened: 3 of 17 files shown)


web/tests/choice-params-with-default.spec.ts
```diff
@@ -4,7 +4,7 @@ import { Workspace } from './lynxkite';
 
 let workspace: Workspace;
 test.beforeAll(async ({ browser }) => {
-  workspace = await Workspace.empty(browser);
+  workspace = await Workspace.empty(await browser.newPage());
```
darabos, line 7: Sorry about all this! I noticed that quite often we have a single test in a file. In this case won't re-use a global object, so we don't need this `newPage()` call. I think passing in a `Page` is a better interface. It allows resetting the test environment inside a test too.

web/tests/data/upload_test.csv
```diff
@@ -0,0 +1,5 @@
+"name"
+"Adam"
+"Eve"
+"Bob"
+"Isolated Joe"
```
darabos, line 5: High-quality test data copied from the original test.

web/tests/drag-select.spec.ts
```diff
@@ -4,26 +4,26 @@ import { Workspace } from './lynxkite';
...
-    });
-    await workspace.addBox({
-      id: 'pr2',
-      name: 'Compute PageRank',
-      x: 100,
-      y: 300,
-      after: 'pr1',
-    });
+  workspace = await Workspace.empty(await browser.newPage());
+  await workspace.addBox({ id: 'ex0', name: 'Create example graph', x: 100, y: 100 });
```
darabos, line 8: Sorry about the whitespace change! We're using 2 spaces for tabs. If this needs to be set in some Lint config, please make the change.

web/tests/upload.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+test.only('can upload and import a simple CSV', async ({ page }) => {
+  const workspace = await Workspace.empty(page);
+  await workspace.addBox({ id: 'imp', name: 'Import CSV', x: 100, y: 100 });
+  const box = await workspace.openBoxEditor('imp');
+  const fileChooserPromise = page.waitForEvent('filechooser');
+  const uploadButton = box.getParameter('filename', '[drop-tooltip="Upload"]');
+  await uploadButton.click();
+  const fileChooser = await fileChooserPromise;
+  await fileChooser.setFiles(resolve('tests/data/upload_test.csv'));
```
darabos, line 14: From https://playwright.dev/docs/api/class-filechooser. Now it really uploads a file. Surprisingly, the original `upload-test.js` does no such thing! It just uses the file that's already on the disk. (This functionality is also tested by `plot.spec.ts` so nothing is lost by fixing this.)

web/tests/choice-params-with-default.spec.ts
lacca0: So do you suggest passing in a Page in files with a single test and syntax like this in files with multiple tests?

web/tests/upload.spec.ts
lacca0: What is a difference between uploading a file and "using the file that's already on the disk"?

web/tests/choice-params-with-default.spec.ts
darabos: > So do you suggest passing in a Page in files with a single test and syntax like this in files with multiple tests?

Yes.

web/tests/upload.spec.ts
darabos: > What is a difference between uploading a file and "using the file that's already on the disk"?

I would hope that the code tells you the difference! To upload a file we click the upload button and select a file. (This code.) To use a file from disk we just enter the filename. (The original code.)

lacca0 (approved):

PR #336, opened 2023-01-25
Convert table-test.js

darabos:
For #238. I've skipped the simple parts, like reading a CSV, because that's already covered in `sql.spec.ts`. But the sorting test is solid. I've also revived two of the tests that were commented out in `table-test.js`. One checks if we can change configuration and re-import. Seems trivial, but all the other tests only import once. I think this is good to check. The other one tests JDBC.

And JDBC sent me on a trip!
- The JDBC test failed with `No suitable driver`. JDBC doesn't work!
- We included several JDBC drivers in `lynxkite.jar`. But they don't get registered. At first I thought it was because of `sbt-assembly`, but no. It's because [`spark-submit` doesn't put our jar on the classpath](https://stackoverflow.com/questions/43257324/what-is-the-use-of-driver-class-path-in-the-spark-command/75231208#75231208)!
- `spark-submit --driver-class-path lynxkite.jar lynxkite.jar` works, but it's ridiculous. I think it's better to leave JDBC driver setup to the user.
- For testing I set up JDBC for SQLite in `with_lk.sh`. I also rewrote this script because 1) it had lot of old stuff and 2) we can rely on the Python API now!
- Turns out starting LynxKite from Python doesn't allow the Python code to terminate! The Akka threads started by Play keep running and blocking the JVM shutdown. I tried a few things. An explicit shutdown via Python's `atexit` is the best I found. It works and doesn't require anything on the user's part.

That's how this PR ended up so complex. Sorry!


python/remote_api/setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='lynxkite-client',
-    version=os.environ.get('VERSION', 'snapshot'),
+    version=os.environ.get('VERSION', '0.1+snapshot'),
```
darabos, line 6: For reference: https://github.com/pypa/setuptools/issues/3772

dependency-licenses/scala.md
```diff
@@ -264,7 +262,6 @@ GPL | [GNU General Public License, Version 3](http://www.gnu.org/licenses/gpl-3.
 GPL | [GNU General Public License, Version 3](http://www.gnu.org/licenses/gpl-3.0-standalone.html) | org.neo4j # neo4j-values # 3.5.2 | <notextile></notextile>
 GPL | [GPL2 w/ CPE](https://www.gnu.org/software/classpath/license.html) | jakarta.transaction # jakarta.transaction-api # 1.3.3 | <notextile></notextile>
 GPL | [GPL2 w/ CPE](https://www.gnu.org/software/classpath/license.html) | jakarta.ws.rs # jakarta.ws.rs-api # 2.1.6 | <notextile></notextile>
-GPL | [The GNU General Public License, v2 with FOSS exception](null) | mysql # mysql-connector-java # 8.0.20 | <notextile></notextile>
```
lacca0, line 267: Why did you delete those?

darabos:
This file is auto-generated based on build.sbt. Removing the MySQL
dependency from build.sbt caused these lines to disappear too.

PR #340, opened 2023-01-26
Convert python-code-generation-test.js

darabos:
For #238.


lacca0 (approved):

darabos:
I had to retry the frontend test. One flaky point to `segmentation.spec.ts:22 › sub-segmentation can be opened`.

image

From the trace I was unable to figure out anything. There are 8 traces for this test and the first 7 are basically empty. I see that connecting the boxes fails, but I can't tell why. I'll take a closer look if it keeps happening.

lacca0 (approved):

PR #341, opened 2023-01-26
Convert project-basics-test.js

darabos:
For #238. I think the tests from `project-basics-test.js` fit in well in `splash.spec.ts`. I've merged them into one test and added an extra check at the end to confirm the view from the directory browser.

This test could start with a new workspace, but instead I've opened "apple". I did this just so we test opening an existing workspace somewhere.


darabos:
I just realized this covers the same feature as `fork-workspace.test.ts`. Which was committed accidentally to `main` with some mistakes that you haven't fixed yet. So I'll just delete `fork-workspace.test.ts` here and save you the work!

lacca0:
> So I'll just delete fork-workspace.test.ts here
Do I need to make a new branch then? In my branch https://github.com/lynxkite/lynxkite/pull/331 several commits are fixes for that file.

darabos:
> > So I'll just delete fork-workspace.test.ts here
> 
> Do I need to make a new branch then? In my branch #331 several commits are fixes for that file.

Oh, I forgot about that. Thanks! You don't need to make a new branch. When these two branches meet, there will be a conflict. The conflict will be easy to resolve by just deleting the file.

In retrospect the right solution would have been to do the fixes in a separate PR, so they don't get held up by unrelated things. But it's not a huge issue.

web/tests/lynxkite.ts
```diff
@@ -1120,25 +1076,15 @@ function sortHistogramValues(values) {
   });
 }
 
-// A promise of the list of error messages.
-function errors() {
-  return $$('.top-alert-message').map(function (e) {
-    return e.getText();
-  });
+export function errors(page: Page) {
```
lacca0, line 1079: So do we need to support it in all the other tests?

lacca0 (approved):

web/tests/lynxkite.ts
darabos: > So do we need to support it in all the other tests?

You mean like check that there are no errors? I don't think that's necessary. In this test we're checking that there _is_ an error.

PR #342, opened 2023-01-26
Convert popup-close-test.js

darabos:
For #238. The original test only had two popups. I was a bit worried with four a popup would cover up one of the boxes we want to click. I've tried it a bunch of times and it looks like this doesn't happen!


lacca0 (approved):

lacca0:
Awesome!

lacca0 (approved):

PR #344, opened 2023-01-27
Convert custom-box-test.js

darabos:
For #238. This is totally not working yet. These are tricky tests to start with, and they triggered a lot of complex problems.

- Setting workspace parameters. I think the core issue here was that `onBlur()` was missing for this parameter type. Hopefully fixed.
- The output box has no output. So the part in `addBox()` where we wait for the output plug to get a color doesn't work. We can't wait for the box to be saved. But users wouldn't wait anyway! We had this wait because you can't click a box before it's saved. I think I fixed that now. (This is the `mouseMoved` diff.)
- Saving a custom box forces a boxcatalog refresh. No way for the user to know about this, so things should just work while it's taking place. I'm yet to fix this.


darabos:
> * Saving a custom box forces a boxcatalog refresh. No way for the user to know about this, so things should just work while it's taking place. I'm yet to fix this.

I think I misinterpreted something here because I didn't fix it but everything seems fine. (We will see what GitHub Actions has to say about it.)

web/app/scripts/operation/parameters-parameter.js
```diff
@@ -16,6 +17,7 @@ angular.module('biggraph').directive('parametersParameter', function(util) {
       });
       util.deepWatch(scope, 'parameters', function(parameters) {
         scope.model = JSON.stringify(parameters);
+        scope.onBlur();
```
lacca0, line 20: Could you comment on your frontend changes in a descriptive form?

lacca0 (approved):

web/app/scripts/app.js
```diff
@@ -130,7 +130,7 @@ angular.module('biggraph').filter('decimal', function() {
 angular.module('biggraph').filter('id', function() {
   return function(x) {
     if (x === undefined) { return x; }
-    return x.toLowerCase().replace(/[ !?,.]/g, '-');
+    return x.toLowerCase().replace(/[ !?,./]/g, '-');
```
darabos, line 133: The custom box IDs ended up with slashes. It kind of works, but it's easier if we don't have slashes.

web/app/scripts/operation/parameters-parameter.js
darabos: Yes. If you look at all parameter types, they all take this `onBlur` function as a parameter. (Look in `operation-parameters.js`.) This is the mechanism that triggers saving parameter changes to the backend. It was missing for `parameters-parameter` (the workspace parameters in the anchor box).

Fixing this helps with breaking the binding loop here that we discussed earlier. But even so it's not perfect and I ended up writing directly into the model.

My plan is that once all the test conversions are done we go through all these `TODO` comments and have another stab at them.

web/app/scripts/workspace/workspace-drawing-board.js
```diff
@@ -289,11 +292,8 @@ angular.module('biggraph').directive(
...
 
         scope.onMouseUpOnBox = function(box, event) {
-          if (box.isDirty || scope.pulledPlug || scope.selection.isActive()) {
-            return;
-          }
           const leftButton = event.button === 0;
-          if (!leftButton || event.ctrlKey || event.shiftKey) {
+          if (mouseMoved || !leftButton || event.ctrlKey || event.shiftKey) {
             return;
           }
```
darabos, line 298: You're right, this deserves an explanation! When you click on a box, we want to open the box popup. A click is basically a mouseup event. But there are some exceptions when we don't want a popup on mouseup. Most importantly, we don't want a popup if this was some sort of dragging event. Maybe we dragged the box, maybe we dragged the arrow between boxes, maybe we dragged a selection rectangle. These are the three checks in the deleted `if`.

I replaced that `if` with a new mechanism that just checks whether the mouse has moved. It's simpler, and it fixes a corner case: When you start dragging a box, `isDirty` is set, to mark this change for saving to the backend. Once it's saved, `isDirty` is cleared. But Playwright is going to click on the box after finishing the movement while it's still marked as dirty! Happens to users too if the network or the backend is slow. This is not an issue with `mouseMoved` — this flag is cleared on mouseup.

darabos (commented):
Thanks! Some explanations added. Let me know if you have more questions!

PR #345, opened 2023-02-02
Another way to save the report for a failed test

darabos:
I think this works, but let's see on GitHub Actions.

### DIFF
--- a/.github/workflows/earthly-test.yml
+++ b/.github/workflows/earthly-test.yml
@@ -158,4 +158,4 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: frontend-test-results
-        path: results.zip
+        path: playwright-report.zip
--- a/.github/workflows/earthly-test.yml
+++ b/.github/workflows/earthly-test.yml
@@ -152,7 +152,7 @@ jobs:
     - name: Earthly version
       run: earthly --version
     - name: Run frontend test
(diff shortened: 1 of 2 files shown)


darabos:
Oof, the Sphynx failure is some randomness issue.

```
[info] - Segment attributes *** FAILED *** (798 milliseconds)
[info]   0.11111111111111116 was not less than 0.01
-- in Compute hub dominance returned Map(0 -> 0.8888888888888888, 1 -> 0.7142857142857143, 2 -> 1.0)
   instead of Map(0 -> 1.0, 1 -> 0.71, 2 -> 1.0) (0.8888888888888888 != 1.0) (NetworKitTest.scala:10)
[info] - Segmentation metrics *** FAILED *** (56 milliseconds)
[info]   10.0 was not less than 0.01
-- in Compute edge cut of segmentation (21.0 != 31.0) (NetworKitTest.scala:10)
```

 I wonder what caused a change. There's no new upload for `libnetworkit`, but maybe one of its dependencies changed, affecting the random seed? I got to say this is a pretty bad way to test... Any ideas?

darabos:
> Oof, the Sphynx failure is some randomness issue.

Another retry fixed it. If it keeps coming up we can change the test to be less specific. Maybe just check that we're not getting all zeroes.

darabos:
Okay, let me merge this so I can get good reports in other PRs if they fail.

PR #351, opened 2023-02-22
[RESEARCH] SQL generation script prototype

lacca0:
For #348. I left it in a test format, delete prints and coloring before usage

### DIFF
--- a/research/generateSQL.py
+++ b/research/generateSQL.py
@@ -1,0 +1,39 @@
+import os
+import openai
+import cmd
+
+openai.api_key = os.getenv("OPENAI_API_KEY")
+openai.api_key_path = "API_KEY"
+
+class color:
+   PURPLE = '\033[95m'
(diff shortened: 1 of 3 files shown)


lacca0:
image
The usage example, sorry for abusing that friendly chef example:)

darabos:
> The usage example, sorry for abusing that friendly chef exercise:)

That's not a great example — the `edges` table already has an `src_name`/`dst_name` column, ideally the generated SQL should use that instead of running a join. (Check out the [docs](https://lynxkite.com/docs/latest/user-guide/#sql1) for a description of the available tables.)

lacca0:
In my experience Codex (as opposed to GPT-3) uses WHERE more often than JOIN, I'm not sure why. Is it better in any sense?
image

darabos:
Thanks! The spirit of this lives on in #353.

PR #356, opened 2023-03-06
Switch from Gulp to Vite

darabos:
I just needed to upgrade something to fix the broken build. But I thought it's time to switch from Yarn to NPM anyway.

Then NPM was warning me about too many vulnerable packages. Most of those are used in the build! But still, we're depending on 10-year-old unmaintained packages. And for what! Our Gulpfile could be replaced by a simple shell script. Or... what do normal people use these days? Is it Webpack?

Looks like [**Vite**](https://vitejs.dev/) has just become the new hotness. The browsers themselves have evolved, and they all support module imports. Okay, so why not Vite?

AngularJS is not directly supported *(cough)* but it's not too bad.
- Instead of `basic-filesort.js` I declared our dependencies by explicitly importing things.
- Instead of applying `basic-annotate.js` on the fly, I applied its changes to the source. So yeah, every dependency is listed 3 times now. But we're not declaring new dependencies very often, so I think that's okay.
- The templates have to be imported too. I imported them.
- I moved the images to `web/app/public/images`. This way they don't have to be imported.
- The `jsError` mechanism had some issues so I disabled it. I've been itching to do that anyway. (We never look at the server side logs to see frontend errors. But it sometimes leads to extra issues, like in #103.)
- There was some dead multi-user code that caused issues. I've put in a hack for that for now. I'll clean it up in a bit.

I still have to fix the documentation generation. And check if there's anything else meaningful in the Gulpfile.

TODO:
- [x] Fix Asciidoc.
- [x] Check `gen_templates.py`.
- [x] Check 3D view.
- [x] Check Firebase.
- [x] Dynamic import for rarely used components: Three and Vega, maybe Firebase and Markdown-it.


darabos:
Hell yeah! We've finally got [`no-floating-promises`](https://typescript-eslint.io/rules/no-floating-promises/)! 🎉  As ridiculous it is, this depended on the Vite switch. I'm not 100% what blocked it before but it was something like `gulp-eslint` being stuck on an old version.

As fallout from these ESLint changes, I made some changes to `lynxkite.ts` that may impact you, @lacca0. I put the whole VisualizationState code into a big comment. Maybe that will help avoid conflicts and you can uncomment it and fix the lint issues.

web/app/scripts/util/drop-tooltip.js
```diff
@@ -1,10 +1,11 @@
 // A tooltip based on the "Drop" popup library.
-'use strict';
+import Drop from 'tether-drop';
+import '../app';
 
 angular.module('biggraph').service('dropTooltipConfig', function() {
   this.enabled = true;
 });
-angular.module('biggraph').directive('dropTooltip', function(dropTooltipConfig) {
+angular.module('biggraph').directive('dropTooltip', ['dropTooltipConfig', function(dropTooltipConfig) {
```
darabos, line 8: Import `dropTooltipConfig`!

web/app/scripts/util/util.js
```diff
@@ -471,7 +473,10 @@ angular.module('biggraph')
     util.frontendConfig = util.globals.then(g => JSON.parse(g.frontendConfig || '{}'));
 
     util.reloadUser = function() {
-      util.user = util.nocache('/ajax/getUserData');
+      util.user = {
```
darabos, line 476: Let's add a comment.

web/app/styles/help.css
```diff
@@ -150,6 +150,6 @@ code {
     content: '';
     display: block;
     height: 100px;
-    background: url(images/logo.png) center / contain no-repeat #004165 !important; /* Brand color. */
+    background: url(/images/logo.png) center / contain no-repeat #004165 !important; /* Brand color. */
```
darabos, line 153: An absolute URL! This would have problems on demo.lynxkite.com. Let's check what's the right way to handle that.

darabos (commented):
We reviewed it together with @lacca0! Thanks!!!

web/tests/lynxkite.ts
```diff
@@ -450,14 +430,14 @@ class PlotState extends PopupBase {
     for (let i = 0; i < count; ++i) {
       const path = await bars.nth(i).getAttribute('d')!;
```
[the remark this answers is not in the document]
darabos: > Why didn't this exclamation mark work? Anyway, let's delete it since we have one below.

The `!` applied to the Promise returned by `getAttribute()` not to the string we `await`. Good catch! I've removed this.

web/app/scripts/python-code.js
```diff
@@ -1,7 +1,7 @@
 // The modal dialog for generated Python code
-'use strict';
+import './app';
 
-angular.module('biggraph').controller('PythonCodeCtrl', function($scope, $uibModalInstance, code) {
+angular.module('biggraph').controller('PythonCodeCtrl', ['$scope', '$uibModalInstance', 'code', function($scope, $uibModalInstance, code) {
```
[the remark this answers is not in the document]
darabos: > Import `code`!

Turns out it's a parameter passed in from `python-code-generator.js`.

web/app/scripts/report-error.js
```diff
@@ -1,8 +1,10 @@
 // The modal dialog for error reporting.
-'use strict';
+import './app';
+import './util/util';
+import jsyaml from 'js-yaml';
+import ClipboardJS from 'clipboard';
 
-angular.module('biggraph').controller('ReportErrorCtrl', function($scope, $uibModalInstance, alert, util) {
-  /* global jsyaml */
+angular.module('biggraph').controller('ReportErrorCtrl', ['$scope', '$uibModalInstance', 'alert', 'util', function($scope, $uibModalInstance, alert, util) {
```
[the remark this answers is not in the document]
darabos: > Import `alert`!

This comes as a parameter from `util.js`.

web/app/scripts/util/drop-tooltip.js
darabos: > Import `dropTooltipConfig`!

No need. It's defines right above this line. 😅 Reviewing 20,000 lines was maybe too much yesterday.

web/app/styles/help.css
darabos: > An absolute URL! This would have problems on demo.lynxkite.com. Let's check what's the right way to handle that.

Setting Vite's base URL to the empty string makes everything relative! This URL, for example, becomes `url(../images/logo.png)` which seems correct.

darabos:
I've addressed the comments. But then the frontend test failed. In that test we open a histogram, change something and then check the histogram again. The histogram check didn't wait — it just grabbed the (old) values from the histogram and asserted something about them. This highlights how important it is to do the checks the other way: by waiting for the expected values to appear.

Hopefully with the fix it will pass now!

PR #359, opened 2023-03-13
rewrite visualization test

lacca0:
Related to #238


web/tests/lynxkite.ts
```diff
@@ -542,6 +542,50 @@ class Side {
...
   }
 
+  expectOperationScalar(name, text) {
+    const cssSelector = 'value[ref="scalars[\'' + name + '\']"';
+    const valueElement = this.toolbox.locator(cssSelector);
+    expect(valueElement.getText()).toBe(text);
+  }
+
+  toggleSampledVisualization() {
+    this.side.element(by.id('sampled-mode-button')).click();
```
darabos, line 552: ESLint should now warn you about missing `await`s like this.

web/tests/visualization.spec.ts
```diff
@@ -0,0 +1,444 @@
...
+        { color: DEFAULT, icon: 'circle', label: '' },
+        { color: DEFAULT, icon: 'circle', label: '' },
+        { color: DEFAULT, icon: 'circle', label: '' },
+    ]);
+    savedPositions = positions(graph);
+
+    await name.visualizeAs('label');
+    await expect(workspace.page.locator('text=Label: name')).toBeVisible();
+    graph = await visualization.graphData();
+    await expect(graph.edges).toConcur(expectedEdges);
```
darabos, line 102: ESLint should also warn you about unneeded `await`s like this.

web/tests/visualization.spec.ts
```diff
@@ -0,0 +1,444 @@
...
+    await expect(graph.vertices).toConcur([
+        { color: DEFAULT, icon: 'circle', label: '' },
+        { color: DEFAULT, icon: 'circle', label: '' },
+        { color: DEFAULT, icon: 'circle', label: '' },
+    ]);
+    savedPositions = positions(graph);
+
+    await name.visualizeAs('label');
+    await expect(workspace.page.locator('text=Label: name')).toBeVisible();
+    graph = await visualization.graphData();
```
darabos, line 101: I'm afraid it's hard to make this reliable. You change a visualization setting. When do you fetch the graph data? When there is no loading animation? It could be that we're just early and the animation has not started yet! So we wait for the animation first? But it could be that we're late and the animation has already finished!

It's for exactly this reason that I replaced `expect(getHistogramValues())` with `expectHistogramValues()` in https://github.com/lynxkite/lynxkite/pull/356/commits/51c313972dcb912bd682d0c72b24ad380c6d1071. Instead of fetching some data and then making assertions about it, the better option with Playwright is to make assertions directly about the page. Those assertions wait to become true.

That said, this is a lot of tests here. It's a lot of work to fix them all like that. If they seem reliable enough as they are, I'm fine with leaving them as they are.

darabos (approved):
Great if you ask me! 😄 Please take a look at my changes before merging. Thanks!

web/tests/lynxkite.ts
```diff
@@ -975,3 +981,10 @@ expect.extend({
     return { pass: true };
   }
 });
+declare global {
+  namespace PlaywrightTest {
+    interface Matchers<R> {
+      toConcur(expected: any): R;
+    }
+  }
+}
```
lacca0, line 990: Wow. Didn't know we could do that.

web/tests/visualization.spec.ts
```diff
@@ -454,10 +477,31 @@ test('visualization context menu', async () => {
...
   const name = editor.left.vertexAttribute('name');
-  const gender = editor.left.vertexAttribute('gender');
-  const income = editor.left.vertexAttribute('income');
-  const age = editor.left.vertexAttribute('age');
-  const location = editor.left.vertexAttribute('location');
-  const weight = editor.left.edgeAttribute('weight');
-  const comment = editor.left.edgeAttribute('comment');
+
+  await name.visualizeAs('label');
+  await visualization.svg.locator('text=Eve >> .. >> .touch').click();
```
lacca0, line 482: What does it do?

web/tests/lynxkite.ts
darabos: > Wow. Didn't know we could do that.

Well `@typescript-eslint/no-namespace` doesn't think we can do it either. I think this sort of type declaration is supposed to be put in a separate file which is then referenced in `tsconfig.json`. But I like putting this next to the definition of `toConcur`. Plus VS Code already can't find our `tsconfig.json`. So I just disabled that rule. 🤷

web/tests/visualization.spec.ts
darabos: > What does it do?

Find `.touch` within the parent of something that has text `Eve`. Because the text itself is not clickable. The nodes have a dedicated `circle.touch` for interaction.

The `..` syntax is [Playwright shorthand](https://playwright.dev/docs/other-locators#parent-element-locator) for `xpath=..`.

PR #360, opened 2023-03-16
Migrate errors-test.js to Playwright

darabos:
For #238. The original of this test was commented out. I revived it a bit reluctantly. Who cares specifically about error reporting for scalars? I still did it and it immediately uncovered a bug. 🤯 

A bit about the bug: it only happens if you run the test a second time. The first time the graph state is opened, the scalars are uncomputed. Computation starts, you see the spinner, then it fails and you see the error ❌.

But if you reload the page and try again (or re-run the test) the computation has already failed when you open the popup. In `getComputedScalarValue` we go on the "-1" branch. On that branch we call `getFuture` and take the error from the returned future. But `getFuture` will never return a future with an error. If the future has already failed, `getFuture` retries. This means the error is not displayed, and there is a retry running without a progress indicator.

I've added a `retry` parameter to `getFuture` so we can get the existing failure with `retry=false`. An alternative would be to just not use `getFuture` here. Look inside `futures` ourselves.


lacca0 (approved):
Wonderful! Good that we revealed a bug.

PR #361, opened 2023-03-16
Migrate operation-selector-test.js to Playwright

darabos:
For #238. Too easy!


lacca0 (approved):
Good example of KISS methodology!

PR #362, opened 2023-03-16
Convert model-test.js to Playwright

darabos:
For #238. I switched from a CSV import to a Python box to make the test more self-contained.

### DIFF
--- a/web/tests/model.spec.ts
+++ b/web/tests/model.spec.ts
@@ -33,7 +33,7 @@ test('ML with Spark', async ({ page }) => {
   await expect(p.locator('#model-method')).toHaveText('Linear regression');
   await expect(p.locator('#model-label')).toHaveText('age');
   await expect(p.locator('#model-features')).toHaveText('yob');
-  await expect(p.locator('#model-details:visible')).toHaveText(/intercept\s*2015/);
+  await expect(p.locator('#model-details')).toHaveText(/intercept\s*2015/);
   await model.popoff();
   await state.close();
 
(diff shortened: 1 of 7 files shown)


darabos:
Context for the last commit: https://github.com/lynxkite/lynxkite/issues/363

lacca0 (approved):

web/tests/model.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    },
+  });
+  state = await ws.openStateView('train', 'graph');
+  const model = state.left.graphAttribute('age_from_yob');
+  await expect(model.value).toHaveText('Linear regression model predicting age');
+  const p = await model.popup();
+  await expect(p.locator('#model-method')).toHaveText('Linear regression');
+  await expect(p.locator('#model-label')).toHaveText('age');
+  await expect(p.locator('#model-features')).toHaveText('yob');
+  await expect(p.locator('#model-details:visible')).toHaveText(/intercept\s*2015/);
```
lacca0: We could just have different IDs for different elements.

lacca0 (approved):

web/tests/model.spec.ts
darabos: > We could just have different IDs for different elements.

Sorry, don't know why I just swept it under the rug! I think when I saw `ng-show` I thought it was two elements that are semantically one, and we just display one or the other. But that's not the case! The second element has the same ID because it was copy-pasted and we didn't notice the ID!

Thanks. Fixed now.

PR #364, opened 2023-03-17
Migrate help-popups-test.js to Playwright

darabos:
For #238.

### DIFF
--- a/web/tests/help-popups.spec.ts
+++ b/web/tests/help-popups.spec.ts
@@ -1,0 +1,39 @@
+// Tests for the help content outside of the full help page.
+import { test, expect } from '@playwright/test';
+import { Workspace } from './lynxkite';
+
+test('help popup in category browser', async ({page}) => {
+  await Workspace.empty(page);
+  await page.locator('.fa-upload').click();
+  const category = page.locator('.box.green');
+  // Help is not visible.
+  await expect(page.getByText('CSV stands for comma-separated values.')).not.toBeVisible();
+  // Help appears on hover.
+  await category.getByText('Import CSV').locator('#help-button').hover();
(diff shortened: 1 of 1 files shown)


lacca0 (approved):

darabos:
The test failure is not great. It shows that even after 30 seconds, a histogram still appeared to be loading.

image

It's possible that the backend was just slow. I tried locally a bunch of times and all 6 tests in `histogram.spec.ts` finished in 30 seconds total. I don't know what GitHub Actions is like, maybe you can end up with bad CPU congestion for half a minute. We could increase the timeout.

But it's also possible that the backend was already finished or never even started working and the frontend showed the animation by mistake. This is more worrying, but if this is possible, it's the first time I see it happen.

For now I'll increase the timeout to help us tell apart which of these it is.

darabos:
The test failure is due to NetworKit returning something different from expected:

```
[info] - Segment attributes *** FAILED *** (730 milliseconds)
[info]   0.11111111111111116 was not less than 0.01 -- in Compute hub dominance returned Map(0 -> 0.8888888888888888, 1 -> 0.7142857142857143, 2 -> 1.0) instead of Map(0 -> 1.0, 1 -> 0.71, 2 -> 1.0) (0.8888888888888888 != 1.0) (NetworKitTest.scala:10)
[...]
[info] - Segmentation metrics *** FAILED *** (137 milliseconds)
[info]   10.0 was not less than 0.01 -- in Compute edge cut of segmentation (21.0 != 31.0) (NetworKitTest.scala:10)
```

It could be non-determinism or a new version of something. (Not NetworKit — it hasn't been updated in 3 months.) I'm repeating the test to learn more. But the frontend test passed, so I'm also merging this!

PR #365, opened 2023-03-21
Clean up Protractor remains

### LINKED ISSUE #238: Rewrite the frontend tests with Playwright

darabos:
We're all converted to Playwright! 🎉  This resolves #238.

### DIFF
--- a/web/app/scripts/util/util.js
+++ b/web/app/scripts/util/util.js
@@ -5,7 +5,6 @@ import firebase from 'firebase/app';
 import 'firebase/analytics';
 
 angular.module('biggraph').service('environment', function() {
-  this.protractor = false; // If we want to handle tests specially somewhere.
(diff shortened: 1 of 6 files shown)


lacca0:
🎉🎉🎉

lacca0 (approved):

PR #369, opened 2023-03-22
Update Spark to 3.3.2

lacca0:


darabos:
We had the version number in two places, and you only changed one. That's why the tests were failing. It's better not to have it in two places. I deleted `SPARK_VERSION` and every reference to it. (And some other stuff.)

I removed the runtime Spark version check too. This means you can now run LynxKite with the wrong Spark version and probably die from Kryo registration issues. That's not super great. When someone sees a `ClassNotFoundException` for a Spark class they will hopefully interpret it as a Spark version issue. If that doesn't work out, we can revert this.

Issue #370, opened 2023-03-23
Workspace parameters glitchy


darabos:
We tried to set them in the tests as a user would ([`lynxkite.ts:910`](https://github.com/lynxkite/lynxkite/blob/2984c21ae9a0b653d57cda2ca58d87fcf65e44cc/web/tests/lynxkite.ts#L910)) but it doesn't work. I can confirm that it also doesn't work when I use it normally. I type something and some characters go missing when I reopen the popup. It's very broken.

darabos:
Fixed in https://github.com/lynxkite/lynxkite/pull/371/commits/ed420227aeb70e970475f100e227a18705a755ea as far as I can tell.

PR #371, opened 2023-03-23
Dropdown parameter for custom boxes

darabos:
When I said this was a lot of work, I didn't anticipate that I would just abuse the default value setting for this. :blush: 

We could make a more complex solution with a dedicated box on the UI for specifying the choices. But it doesn't sound like a good deal, even from a user's perspective. Simple is better.

image

image

I would love to add a test, but the workspace parameters are pretty broken at the moment. (#370) Let's fix that and add a test for this at the same time.

### DIFF
--- a/web/tests/lynxkite.ts
+++ b/web/tests/lynxkite.ts
@@ -1,5 +1,4 @@
 // Shared testing utilities.
-// TODO: This is being migrated from test-lib.js. We will clean it up at the end.
 import { expect, Locator, Page } from '@playwright/test';
 
 // Mirrors the "id" filter.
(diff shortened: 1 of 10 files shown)


web/tests/lynxkite.ts
```diff
@@ -1,5 +1,4 @@
 // Shared testing utilities.
-// TODO: This is being migrated from test-lib.js. We will clean it up at the end.
```
darabos, line 2: It's not too bad now! I'm not sure all the methods are actually used but I couldn't find a way to find dead code. There are two TODOs left. Both are reasonable, but not urgent.

Issue #375, opened 2023-03-28
Eager Python box


darabos:
The Python box has to declare its outputs, because we want to build on the outputs even before it's executed. That's LynxKite for ya. But we could add a version of the box that is like the import boxes. You would have to trigger it with a button. But you wouldn't have to declare the output!

This is especially important when the Python box is importing or enriching data from an external source, such as WikiData. We don't know ahead of time what sort of columns we're getting.

PR #376, opened 2023-03-28
A tool for fetching extra data from WikiData

darabos:
I thought this was a great idea, but then I realized we need to do #375 before this can be really useful. I think we can still merge it because it doesn't harm anyone and I'm going to use it for a demo.

What it does is it looks up each item on WikiData, tries to pick the right match, and fetches a bunch of properties that look useful. For example, for the three cities you get:

`head of government`  `population` `twinned administrative body`                                      `country`     `significant event`  `male population`  `female population` `located in the administrative territorial entity`  `social media followers`

For the four actors:
`name` `award received` `unmarried partner` `spouse`                   `nominated for`    `net worth`                                    `member of`  `social media followers`

When you ask for Mississippi and other states, you get capitals and stuff. If you ask for Mississippi and other rivers, you get their length, "mouth of watercourse", etc. But if Amazon is listed with other companies, you get industries and CEOs.

If you mix actors and cities, you will end up with some mix of columns. It's not too bad.

### DIFF
--- a/sphynx/python/wikidata.py
+++ b/sphynx/python/wikidata.py
@@ -1,0 +1,192 @@
+'''Utility functions that use WikiData. To be used from 'Compute in Python' boxes.'''
+import dataclasses
+import httpx
+import itertools
+from pprint import pprint
+import shelve
+import tqdm
+
+# LynxKite may run this code multiple times, so we cache the results.
+shelf = shelve.open('wd.cache')
+
+
+def get(url, **kwargs):
+  '''Sends a GET request to the given URL, caching the result. Returns a JSON object.'''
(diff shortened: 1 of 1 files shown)


darabos:
I love the data in WikiData, but this is too slow. It takes hours for a few thousand rows.

If we want to offer this feature, I think a better option will be to take the WikiData dump and build our own server that can efficiently support this sort of lookup.

That way we could do a lot of optimizations. We could drop external-id properties. We could resolve labels ahead of time. We could collect stats about entities and properties and help that for picking the right ones. (E.g. the more commonly used property is probably better.) We could add reverse properties. (Like a book has an author, but I don't think the author has a property for each book.)

PR #378, opened 2023-04-04
Remove personal data from license file

lacca0:


darabos (approved):
Thanks!

(Now I understand why you complained about having both Makefile and Earthfile...)

PR #379, opened 2023-04-05
OpenAI box fixes

darabos:
Sorry, looks like I forgot to update the builtin box in #368.

But if you don't mind, I'll fix two more issues here:
- The custom box description is not rendered as Markdown in the box popup. (As we noticed yesterday.)
- I'll display the table schema and sample column by column.

PR #381, opened 2023-04-11
Fix Python test to be compatible with Pandas 2.0.0

darabos:
Fixes the failure seen on #379:

```
ValueError: Multi-dimensional indexing (e.g. `obj[:, None]`) is no longer supported.
Convert to a numpy array before indexing instead.
```

### DIFF
--- a/conda-env.yml
+++ b/conda-env.yml
@@ -52,7 +52,7 @@ dependencies:
 # For R API and tools.
 - r
 - r-essentials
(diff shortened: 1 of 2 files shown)


darabos:
Uh, now R has something to say too.

```
This build of the arrow package does not support Datasets
```

Why not!

lacca0 (approved):

PR #387, opened 2023-04-12
Update PartitionedParquet code for SPARK-40113

darabos:
This is @lacca0's change from another repo. [SPARK-40113](https://github.com/apache/spark/commit/cee30cb4994b178f6999cbbefdcaac21f0d60915) is coming in Spark 3.4.0 but it's already part of the Databricks runtime. I don't think there's any harm in this change. When used with Spark 3.3.0 it may break some pushed filters, but I don't think we rely on those. At least the tests pass!

(Still, let's not merge it before the release.)


darabos:
Let's merge it so we get the most experience with it before the next release.

PR #397, opened 2023-04-19
Cut 5.3.0 in changelog

darabos:
It's not a user-visible change, so it doesn't really matter. But if we're mentioning Vite, we should mention Playwright too.


darabos:
Oops! I didn't realize 5.3.0 was not in the changelog yet! Okay, then that's the main point of this PR now, and adding Playwright is a minor thing.

PR #402, opened 2023-06-01
Update hadoop-aws to 3.3.4

darabos:
Fixes the baffling `For input string: "64M"` error message when trying to read from S3 in LynxKite 5.3.0.


darabos:
The test failure is not great. `sphynx-build` fails with no error. Locally I had SBT run out of memory. What's up with this build?! 💀

darabos:
The same tests failed, but for completely different reasons!
- Backend test with Sphynx: Two tests in NetworKitTest failed because they return unpredictable results.
- Python API test: `sbt` failed with `OutOfMemoryError: Java heap space`.
- Frontend test: `apt-get install -y chromium-browser` failed with `E: Failed to fetch http://archive.ubuntu.com/ubuntu/dists/jammy-updates/main/binary-amd64/Packages.gz  Hash Sum mismatch`.

😭 I'll just retry.

PR #407, opened 2023-06-12
Find conda packages instead of anaconda ones

lacca0:


darabos (approved):
Not "instead of", but in addition to! Thanks!

darabos:
The build failure is not your fault:
```
earthly/buildkitd:v0.6.30: exit status 125: Unable to find image 'earthly/buildkitd:v0.6.30' locally
docker: Error response from daemon: Get "https://registry-1.docker.io/v2/": net/http: request canceled (Client.Timeout exceeded while awaiting headers).
```

That said, let's re-run it to make sure this doesn't break the CI build.

PR #410, opened 2023-07-10
Allow KITE_DRAWING_OVERALL to take effect

darabos:
Turns out `KITE_DRAWING_OVERALL` was setting a default visualization size limit, but the frontend was overriding it with a fixed 10,000. 😮‍💨 This change removes the special casing for the 3D view, but I think that's not too bad. You can configure the limit after all!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Please add changes to "master", preferably ordered by their significance. (Most 
 
 ### master
 
+- `KITE_DRAWING_OVERALL` now properly controls the limits on the size of visualizations. [#410](https://github.com/lynxkite/lynxkite/pull/410)
+
 ### 5.3.1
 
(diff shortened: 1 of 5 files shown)


darabos:
The tests fail before even starting. Hoping for an intermittent issue with Conda infrastructure.

```
               +base | --> RUN ./conda-env.sh build > env.yml && micromamba install -y -n base -f env.yml
               +base | error    libmamba ZSTD decompression error: Unknown frame descriptor
               +base | error    libmamba ZSTD decompression error: Unknown frame descriptor
               +base | critical libmamba Multiple errors occured:
               +base |     Download error (23) Failed writing received data to disk/application [https://conda.anaconda.org/r/noarch/repodata.json.zst]
               +base |     Failure writing output to destination
               +base |     Subdir pyg/noarch not loaded!
               +base |     Subdir conda-forge/noarch not loaded!
               +base |     Subdir r/noarch not loaded!
```

PR #414, opened 2024-03-08
New box for text embedding

darabos:


darabos:
I had to update some Python dependencies, but I did not update `conda-env.yml` yet. I'll do that and run some tests.

The box in action:
zh_2200_ (9)

It's pretty nice that you can compare different embeddings with just a click.

doragera (approved):
I am still learning Lynxkite, but looks good to me!

darabos:
I spent way too much time trying to get the test working in CI but I can't.

`openjdk 11.0.15` requires `jpeg >=9e,<10a`. But `sentence-transformers 2.6.1` requires `torchvision 0.16.1` which requires `libjpeg-turbo >=3.0.0,<4.0a0` which requires `jpeg <0.0.0a`. I don't know why they care about the `jpeg` package so much. We're not using Torchvision at all. The available `jpeg` versions are `9e`, `9d`, `9c`, and `8d`. I'm lost.

Anyway, even though `thenlper/gte-small` is a small (30M parameter) model, I don't feel great about downloading it on every pull request. (Even if we're downloading 2 GB of Conda dependencies already.)

PR #415, opened 2024-04-08
Skip color legend if long, fix CI issues

darabos:
I created CSS color strings in a Python box for the visualization in https://github.com/lynxkite/lynxkite/pull/414. The colors then end up in the legend, making it super long with 100+ nodes. So I added a cutoff. It looks nice:
image
(Here the color is based on the first three dimensions of the embedding, while the first two are used as for layout.)

### DIFF
--- a/.github/workflows/earthly-test.yml
+++ b/.github/workflows/earthly-test.yml
@@ -125,21 +125,15 @@ jobs:
     env:
       FORCE_COLOR: 1
     steps:
-    - name: Maximize build space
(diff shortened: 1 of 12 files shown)


darabos:
Made some progress fixing the tests. But:

```
 +frontend-test-save |     Error: ENOSPC: no space left on device, write
```

😭

darabos:
At last:
image

In the end I think I made some frontend tests less flaky and made the Python requirements more precise. Also updated a few dependencies for no particular reason.

PR #421, opened 2024-05-28
Add missing importFooNow functions in Python API

darabos:
Two operations were missing the "Now" functions: `importFromBigQueryRawTable` and `importFromBigQueryStandardSQLResult`. Missing "Now" functions are a problem when working with these sources in Python. Having unnecessary "Now" functions is less problematic. So I changed the code to cover all future operations starting with "import".

I also found a test that was broken in a funny way that still let it pass.

### DIFF
--- a/python/remote_api/src/lynx/kite.py
+++ b/python/remote_api/src/lynx/kite.py
@@ -34,9 +34,8 @@ from collections import deque, defaultdict, Counter
 from typing import (Dict, List, Union, Callable, Any, Tuple, Iterable, Set, NewType, Iterator,
                     TypeVar, Optional, Collection, TYPE_CHECKING)
 import requests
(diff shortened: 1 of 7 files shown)


python/remote_api/src/lynx/kite.py
```diff
@@ -1161,7 +1145,7 @@ def _fn_id(fn: Callable):
 
 
 def _is_lambda(f: Callable):
-  def LAMBDA(): return 0
+  LAMBDA = lambda: 0 # fmt: off (autopep8 wants to turn this into a function.)
```
darabos, line 1148: So the original author (@erbenpeter) wrote this correctly, but the pre-commit hook turned it into the silly version automatically. :smile:

